PR #2, opened 2019-02-20
Add a plugin to check platform name

pnuu:
This PR adds a plugin that aborts processing of the scene if the platrom name isn't in the list of processed platforms.

### DIFF
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -29,7 +29,7 @@ import sys
 from multiprocessing import Process
 import yaml
 import time
-from trollflow2 import gen_dict_extract
+from trollflow2 import gen_dict_extract, AbortProcessing
 from collections import OrderedDict
 from six.moves.urllib.parse import urlparse
 
@@ -78,7 +78,10 @@ def process(msg, prod_list):
     job = message_to_job(msg, config)
     for wrk in config['workers']:
         cwrk = wrk.copy()
-        cwrk.pop('fun')(job, **cwrk)
+        try:
+            cwrk.pop('fun')(job, **cwrk)
+        except AbortProcessing:
+            LOG.info("Processing aborted for this scene")
 
 
 def main():
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -79,9 +79,10 @@ def process(msg, prod_list):
     for wrk in config['workers']:
         cwrk = wrk.copy()
         try:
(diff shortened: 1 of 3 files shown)


bin/satpy_launcher.py
```diff
@@ -78,7 +78,10 @@ def process(msg, prod_list):
     job = message_to_job(msg, config)
     for wrk in config['workers']:
         cwrk = wrk.copy()
-        cwrk.pop('fun')(job, **cwrk)
+        try:
+            cwrk.pop('fun')(job, **cwrk)
+        except AbortProcessing:
+            LOG.info("Processing aborted for this scene")
```
mraspaud: Can we add the error message to the info ?

pnuu: In think the plugin should handle that, it has more info on what happened. Although, I'll add the name of the plugin to the message.

bin/satpy_launcher.py
```diff
@@ -79,9 +79,10 @@ def process(msg, prod_list):
     for wrk in config['workers']:
         cwrk = wrk.copy()
         try:
-            cwrk.pop('fun')(job, **cwrk)
+            func = wrk('fun')
+            func(job, **cwrk)
```
mraspaud: how does this work if `fun` isn't removed from `cwrk` ?

bin/satpy_launcher.py
```diff
@@ -78,7 +78,11 @@ def process(msg, prod_list):
     job = message_to_job(msg, config)
     for wrk in config['workers']:
         cwrk = wrk.copy()
-        cwrk.pop('fun')(job, **cwrk)
+        try:
+            func = wrk('fun')
+            func(job, **cwrk)
+        except AbortProcessing:
+            LOG.info("Processing aborted by '%s'", func.__name__)
```
mraspaud: don't we want to print the message too, ? like
```suggestion
            LOG.info("Processing aborted by '%s': %s", func.__name__, str(err))
```

bin/satpy_launcher.py
pnuu: I hoped it would? I'll add the `pop` in there.

bin/satpy_launcher.py
pnuu: The plugin should already have logged the reason, so don't think duplicate message is needed.

bin/satpy_launcher.py
pnuu: `pop` added

PR #5, opened 2019-02-21
Add possibility to prioritize production by area

pnuu:
This PR makes it possible to prioritize areas to be processed in certain order by giving them priority value. If priorities are not given, all defined areas are processed together. All areas with the same priority are processed together.

### DIFF
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -70,6 +70,8 @@ def get_area_priorities(product_list):
     plist = product_list['product_list']
     for area in plist.keys():
         prio = plist[area].get('priority', DEFAULT_PRIORITY)
+        if prio is None:
+            prio = DEFAULT_PRIORITY
         if prio not in priorities:
             priorities[prio] = [area]
         else:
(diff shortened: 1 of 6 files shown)


trollflow2/launcher.py
```diff
@@ -0,0 +1,114 @@
...
+def message_to_jobs(msg, product_list):
+    jobs = OrderedDict()
+    priorities = get_area_priorities(product_list)
+    # TODO: check the uri is accessible from the current host.
+    input_filenames = [urlparse(uri).path for uri in gen_dict_extract(msg.data, 'uri')]
+    for prio, areas in priorities.items():
+        jobs[prio] = OrderedDict()
+        jobs[prio]['input_filenames'] = input_filenames.copy()
+        jobs[prio]['input_mda'] = msg.data.copy()
+        jobs[prio]['product_list'] = {}
```
mraspaud: Shouldn't we add the `common` section too ?

pnuu: The `common` section is inside the `product_list` on the same level as `products`.

mraspaud: I don't see it, where does this happen ?

mraspaud: ah, sorry, got it

PR #15, opened 2019-03-01
Fix plist_iter to provide the area and product keys too

mraspaud:
This allows the correct usage of keys vs names for areas and products

### DIFF
--- a/trollflow2/__init__.py
+++ b/trollflow2/__init__.py
@@ -119,14 +119,14 @@ def save_datasets(job):
     base_config.pop('dataset', None)
     for fmat, fmat_config in plist_iter(job['product_list']['product_list'], base_config):
         fname_pattern = fmat['fname_pattern']
-        outdir = fmat['output_dir']
-        filename = compose(os.path.join(outdir, fname_pattern), fmat)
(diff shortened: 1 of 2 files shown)


trollflow2/__init__.py
```diff
@@ -122,10 +123,10 @@ def save_datasets(job):
         filename = compose(os.path.join(outdir, fname_pattern), fmat)
         fmat.pop('format', None)
         try:
-            objs.append(scns[fmat['areaname']].save_dataset(fmat['productname'], filename=filename, compute=False, **fmat))
+            objs.append(scns[fmat['area']].save_dataset(fmat['product'], filename=filename, compute=False, **fmat))
         except KeyError as err:
             LOG.info('Skipping %s: %s', fmat['productname'], str(err))
-        fmat_config['filename'] = filename
```
pnuu: Is this added intendation intentional? It removes the filenames from the product list `formats` list.

pnuu:
I updated the tests to match the data. It seems that the `filename` isn't added to the product list formats anymore, and I added a comment pointing to that. If not, revert my last commit to the tests and reindent :-)

PR #16, opened 2019-03-08
Make it possible to delay composite creation

pnuu:
This PR makes it possible to delay composite creation until after resampling. In most cases this is much faster, and is thus is now the default behaviour. To create composites _before_ resampling the user can set `delay_composites: False` in the `common` section of the production config file.

As an example, 5-area 55-product product list with bilinear interpolation and pre-computed resampling LUTs the processing times (with a single core) are:
- `delay_composites: False`: 35 minutes
- `delay_composites: True`: 5 minutes

### DIFF
--- a/examples/pl.yaml
+++ b/examples/pl.yaml
@@ -23,6 +23,10 @@ common:
   #    avhrr/3: avhrr-3
   #sunzen_check_lon: 25.0
   #sunzen_check_lat: 60.0
+  # Set delay_composites to `False` to create composites before resampling.
+  #   Default is to create them after resampling (`True`), which is much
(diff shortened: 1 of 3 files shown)


pnuu:
And timing for `nearest` resampling for the same product list:
- `delay_composites: False`: 13.5 minutes
- `delay_composites: True`: 4 minutes

pnuu:
I renamed the parameter just so that it would be much less confusing for the user. And the parameter name `generate` is too generic to be as-is in the configuration file, IMHO.

mraspaud:
Should we propose a patch to satpy then ?

PR #20, opened 2019-03-14
Handling nicer the situation where the scene cannot be created when th…

adybbroe:
…ere is no reader for the file type


adybbroe:
Ok, I think I see what you mean. I will make an issue and close/merge this as is. Hope that is okay!?

PR #22, opened 2019-03-20
Make publish topic composable

pnuu:
This PR makes it possible to use `trollsift` patterns in the publisher topics. All the keys in passed to the message can be used.


pnuu:
This PR also converts or `yaml.load()` calls in tests to use `yaml.load(..., loader=UnsafeLoader)`.

PR #25, opened 2019-03-22
Add a possibility to send emails about crashes

pnuu:
As Trollflow2 handles crashes internally, `supervisord` don't report anything about them, and only trace of the problem is in the log files. This PR adds a possibility to send emails with the error message.

### DIFF
--- a/trollflow2/__init__.py
+++ b/trollflow2/__init__.py
@@ -129,7 +129,7 @@ def save_datasets(job):
         filename = compose(os.path.join(fmat['output_dir'], fname_pattern), fmat)
         fmat.pop('format', None)
         try:
-            objs.append(scns[fmat['area']].save_dataset(fmat['product'], filename=filename, compute=False, **fmat))
+            objs.append(scns[fmat['area']].save_dataset(fmat['product'], compute=False, **fmat))
         except KeyError as err:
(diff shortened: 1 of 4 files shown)


pnuu:
The functionality is now more generic and user can supply their own crash handler.

pnuu:
I think I'll still add a possibility to collect info from `input_mda` to be passed to the handlers.

trollflow2/__init__.py
```diff
@@ -128,8 +128,11 @@ def save_datasets(job):
         fname_pattern = fmat['fname_pattern']
         filename = compose(os.path.join(fmat['output_dir'], fname_pattern), fmat)
         fmat.pop('format', None)
+        fmat.pop('filename', None)
```
mraspaud: Shouldn't it raise an error instead ? there shouldn't be a 'filename' there, right ?

pnuu: Yeah, I was wondering how the filename ends up in the format, but thought it was by design as it was there always for the 2nd and later composites. It is added in the `else` branch, and as I don't understand what `plist_iter` is doing, thought it somehow propagated to the later products.

mraspaud: I'm afraid that could be a sign that the format section is shared between multiple products.

PR #27, opened 2019-03-26
Add overviews to output images

pnuu:
This PR adds a feature for embedding "overviews" to the saved TIFF images.


trollflow2/__init__.py
```diff
@@ -301,6 +305,29 @@ def sza_check(job):
             dpath.util.delete(product_list, '/product_list/%s' % area)
 
 
+def add_overviews(job):
+    """Add overviews to images already written to disk."""
+    # Get the formats, including filenames and overview settings
+    product_list = job['product_list']['product_list']
+    for area in product_list:
```
mraspaud: Why not use plist_iter ?

PR #29, opened 2019-04-02
Add non-noisy publisher

pnuu:
This PR adds a publisher that doesn't broadcast its own name and address.

Also renames the `FilePublisher` as `NoisyFilePublisher`.


pnuu:
This turned out to be obsolete. The problem it was trying to solve was somewhere else.

PR #32, opened 2019-04-08
Feature sunlight coverage

mraspaud:
This PR adds a plugin to check for sunlight coverage over areas.


trollflow2/__init__.py
```diff
@@ -314,6 +314,53 @@ def sza_check(job):
...
+    from pyresample.boundary import AreaDefBoundary
+    from trollsched.spherical import get_twilight_poly
+
+    adp = AreaDefBoundary(area_def, frequency=100)
+    poly = get_twilight_poly(start_time)
+
+    daylight = adp.contour_poly.intersection(poly)
+    if daylight is None:
+        if sun_zenith_angle(start_time, *area_def.get_lonlat(0, 0)) < 90:
+            return 1
```
pnuu: Should the return values be floats?

trollflow2/__init__.py
```diff
@@ -314,6 +314,53 @@ def sza_check(job):
...
+    """Get the sunlight coverage of *area_def* at *start_time*."""
+    from pyresample.boundary import AreaDefBoundary
+    from trollsched.spherical import get_twilight_poly
+
+    adp = AreaDefBoundary(area_def, frequency=100)
+    poly = get_twilight_poly(start_time)
+
+    daylight = adp.contour_poly.intersection(poly)
+    if daylight is None:
+        if sun_zenith_angle(start_time, *area_def.get_lonlat(0, 0)) < 90:
```
pnuu: Is check in one corner of the area enough?

trollflow2/__init__.py
```diff
@@ -314,6 +314,53 @@ def sza_check(job):
             dpath.util.delete(product_list, '/product_list/%s' % area)
 
 
+def check_sunlight_coverage(job):
+    """Remove products with too low daytime coverage.
+
+    This plugins looks for a parameter called `min_sunlight_coverage` in the
+    product list, expressed in % (so between 0 and 100). Everything under it not
+    matching this sunlight coverage will be discarded.
```
pnuu: The last sentence needs some rewording to be more understandable. Maybe something like `If the sunlit fraction is less than configured, the affected products will be discarder.`?

setup.py
```diff
@@ -51,7 +51,7 @@
       data_files=[],
       zip_safe=False,
       install_requires=install_requires,
-      tests_require=['mock'],
+      tests_require=['mock', 'pytroll-schedule', 'pyresample'],
```
pnuu: Do we need these packages for the tests or could (should?) they be mocked?

trollflow2/__init__.py
mraspaud: I can do that

trollflow2/__init__.py
mraspaud: in principle yes as, having an empty intersection means either full light, or full night

setup.py
mraspaud: I can mock them if you prefer

trollflow2/__init__.py
mraspaud: ok

PR #33, opened 2019-04-09
Fix hanging publisher

pnuu:
In the case when there were a crash or all the products were discarded the later plugins were not called. When `FilePublisher` was used, this caused the initialised publisher not to be closed, as the publisher was stopped in the `__call__()` method. With this PR the publisher will be closed when the instance is deleted and garbage collected.

### DIFF
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -145,13 +145,16 @@ def process(msg, prod_list):
                     cwrk = wrk.copy()
                     cwrk.pop('fun')(job, **cwrk)
             except AbortProcessing as err:
-                LOG.info(str(err))
+                LOG.info("Worker %s failed: %s", str(wrk['fun']), str(err))
     except Exception:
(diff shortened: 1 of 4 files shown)


trollflow2/launcher.py
```diff
@@ -145,14 +144,17 @@ def process(msg, prod_list):
...
+                LOG.info("Worker %s failed: %s", str(wrk['fun']), str(err))
     except Exception:
         LOG.exception("Process crashed")
         if "crash_handlers" in config:
             trace = traceback.format_exc()
             for hand in config['crash_handlers']['handlers']:
                 hand['fun'](config['crash_handlers']['config'], trace)
 
+    # Remove config so all remaining references are removed
+    del config
```
mraspaud: This will fail if the call to `open` crashed

pnuu: Ah, good catch. I'll wrap it in try/except.

PR #35, opened 2019-04-26
Add pass coverage computation to sunlight coverage checking

mraspaud:
In order to get a realistic daylight coverage estimate of the data, not only the sunlight coverage of the data must be computed, but also the coverage of the pass/actual data. This patch implements this.


adybbroe:
@mraspaud Is there a reason why this is not yet merged!?

adybbroe:
@mraspaud See above

PR #38, opened 2019-05-28
Handle gracefully the situation when a dataset is not loaded

adybbroe:
Handle the situation when a product/dataset is not loaded due to insufficient spectral bands available

### DIFF
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -125,6 +125,21 @@ product_list:
             writer: geotiff
 """
 
+yaml_test3 = """common:
+  something: foo
+  min_coverage: 5.0
+product_list:
+  euron1:
+    areaname: euron1_in_fname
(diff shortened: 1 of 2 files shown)


adybbroe:
@mraspaud I added a unittest, I guess it could be better, but it fails if I don't have the try-except. I might see if I can improve it tomorrow.

trollflow2/tests/test_trollflow2.py
```diff
@@ -648,6 +663,41 @@ def test_sza_check(self, sun_zenith_angle):
         self.assertFalse('germ' in job['product_list']['product_list'])
 
 
+class TestCheckSunlightCoverage(unittest.TestCase):
```
mraspaud: I think you forgot to add the new class to the test suite

adybbroe:
Yes, your are right, thank you @mraspaud 
Now, fixed!

PR #39, opened 2019-06-13
Refactor trollflow2 to put plugins in their own file

mraspaud:
This PR refactors trollflow2 to take the contents out of the main __init__.py, which was causing problems for testing and ugly code.

WARNING: All the existing product lists will break as the plugins are now in `trollflow2.plugins`, not `trollflow2`.


trollflow2/dict_tools.py
```diff
@@ -0,0 +1,85 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>
+
+# Workaround for unittests so that satpy and posttroll installations
+# are not necessary
+"""Tools for product list operations."""
+
+import dpath
+
+def plist_iter(product_list, base_mda=None, level=None):
```
pnuu: Some explanation on the workings of `plist_iter()` would be nice. Docstring at the least.

trollflow2/dict_tools.py
```diff
@@ -0,0 +1,85 @@
...
+                yield pconfig, prod_config
+                continue
+            for idx, file_config in enumerate(pconfig.get('formats', [{'format': 'tif', 'writer': 'geotiff'}])):
+                fconfig = pconfig.copy()
+                fconfig.pop('formats', None)
+                fconfig.update(file_config)
+                yield fconfig, file_config
+
+
+def gen_dict_extract(var, key):
```
pnuu: Docstring missing.

trollflow2/tests/test_launcher.py
```diff
@@ -180,117 +178,116 @@ def test_message_to_jobs_minimal(self):
...
-            run(prod_list, topics=['/topic3'])
-        except KeyboardInterrupt:
-            pass
-        lc_.assert_called_with(topics=['/topic3'])
-
-    @mock.patch('trollflow2.launcher.yaml.load')
-    @mock.patch('trollflow2.launcher.open')
-    @mock.patch('trollflow2.launcher.ListenerContainer')
-    def test_run_keyboard_interrupt(self, lc_, _open, yaml_load):
+        with mock.patch('trollflow2.launcher.yaml.load') as yaml_load,\
```
pnuu: Are these context manager versions somehow better than having the decorators? Or is it required as part of the new `TestCase` wrapper class? I think it just adds one unnecessary level of intendation for a long piece of code and creates unnecessarily long diffs.

trollflow2/tests/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        else:
+            import sys
+            del sys.modules['trollflow2.plugins']
+            break
+    if module_patcher is not None:
+        module_patcher.stop()
+    return module_patcher
+
+
+module_patcher = find_missing_modules()
```
pnuu: This could be inside the `TestCase.setUp()` method, no need as a module-wide variable.

trollflow2/tests/test_launcher.py
mraspaud: yes, it's not the nicest, but since we import plugins in the function, we need to patch the function after the import

trollflow2/tests/utils.py
mraspaud: if we do that inside the setup, we need to search for missing modules before every test

trollflow2/tests/test_launcher.py
mraspaud: and yes, also the patcher needs to be in place, so the patches need to be done inside the test function

trollflow2/tests/utils.py
pnuu: Ah, good point.

trollflow2/dict_tools.py
mraspaud: done

trollflow2/dict_tools.py
mraspaud: fixed

pnuu:
LGTM

PR #41, opened 2019-06-18
Allow saving files to a temporary name and small fixes

mraspaud:
Introduce the `use_tmp_file` parameter to allow saving the data to a temp file before renaming it to it's final name.

We also included fixes for the previous PR #40

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -21,16 +21,23 @@
 
 """Trollflow2 plugins."""
 
+import os
+from logging import getLogger
+from tempfile import NamedTemporaryFile
+
+import dpath
+import rasterio
 from posttroll.message import Message
 from posttroll.publisher import NoisyPublisher
-from satpy import Scene
-from satpy.writers import compute_writer_results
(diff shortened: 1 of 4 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -133,23 +130,36 @@ def save_datasets(job):
...
     base_config = job['input_mda'].copy()
     base_config.pop('dataset', None)
+    renames = {}
     for fmat, fmat_config in plist_iter(job['product_list']['product_list'], base_config):
         fname_pattern = fmat['fname_pattern']
         filename = compose(os.path.join(fmat['output_dir'], fname_pattern), fmat)
+        directory = fmat['output_dir']
+        if not os.path.exists(directory):
+            os.makedirs(directory)
+        if fmat.get('use_tmp_file', False):
```
pnuu: Short explanation here would be nice: what's the overall purpose of the temporary file. Something like "Save file to a temporary file and rename after writing so that partly written files will not be accessed"

mraspaud: thanks, I'll fix that

PR #42, opened 2019-06-19
Add option to run in test-mode providing a specific message on the co…

adybbroe:
…mmand line


adybbroe:
Hey @mraspaud Thought you had merged this one?
What happened here? I don't see it in the list of open PRs neither in the list of closed ones???

trollflow2/launcher.py
```diff
@@ -54,26 +54,52 @@
...
+            msg = fpt.readline().strip('\n')
+
+    return msg
+
+
+def run(prod_list, topics=None, test_message=None):
+
+    tmessage = get_test_message(test_message)
+    if tmessage:
+        from threading import Thread as Process
```
pnuu: Any specific reason to use `Threading` instead of `multiprocessing` for "offline" mode?

trollflow2/launcher.py
```diff
@@ -54,26 +54,52 @@
...
 
-    listener = ListenerContainer(topics=topics)
+    if not tmessage:
+        listener = ListenerContainer(topics=topics)
 
     while True:
         try:
-            msg = listener.output_queue.get(True, 5)
+            if tmessage:
+                mymessage = tmessage
```
pnuu: This is unnecessary, `tmessage` can be used directly without creating another variable.

bin/satpy_launcher.py
```diff
@@ -31,19 +32,23 @@
...
-
 if __name__ == "__main__":
-    main()
+
+    parser = argparse.ArgumentParser(
+        description='Launch trollflow2 processing with Satpy listening on the specified Posttroll topic(s)')
+    parser.add_argument("topics", nargs='*',
+                        help="The topic(s) - a space separated list of topics",
+                        type=str)
+    parser.add_argument("--product_list", '-p',
```
pnuu: I think product list filename should be kept as positional argument (without a command-line switch) so that command for operations would be a bit cleaner and wouldn't need to be changed.

trollflow2/launcher.py
adybbroe: Otherwise I can't use pdb

trollflow2/launcher.py
adybbroe: Right, thx, fixed.

bin/satpy_launcher.py
adybbroe: I don't really agree. Also, because I am not sure how to do that in a consistent/safe way?

pnuu:
```python
parser.add_argument("config", help="config file to be used")```

PR #46, opened 2019-07-04
Make writing via tmp file more robust

mraspaud:
Ensure empty tmp files aren't kept and that only valid files are renamed.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -1,5 +1,7 @@
 linters:
   flake8:
     fixer: true
+    python: 3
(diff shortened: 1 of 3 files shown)


pnuu:
Is this forcing the use of temporary files when saving, or can it be disabled?

mraspaud:
It's optional and disabled by default: https://github.com/pytroll/trollflow2/pull/46/files#diff-13b03419007f554e7e792153d39ab531R162-R169

PR #55, opened 2019-10-03
Make productname, areaname and format optional in the published message

mraspaud:


mraspaud:
product and area will always be there, and the rest is optional as far as I can see, so we should be fine.

PR #59, opened 2019-10-11
Add "nameserver" and "addresses" command-line options

pnuu:
This PR adds two command-line options so that `nameserver` and `addresses` can be passed to `ListenerContainer`. With these options it should be possible to run `trollflow2` without multicast.

This also adds `nameservers` kwarg to `FilePublisher` class.

 - [x] Tests passed 
 - [x] Tests updated
 - [x] Fully documented

### DIFF
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -46,10 +46,11 @@ def parse_args():
                         help="Log config file (yaml) to use",
                         type=str, required=False)
     parser.add_argument('-n', "--nameserver", required=False, type=str,
-                        help="Nameserver to connect to")
+                        help="Nameserver to connect to", default='localhost')
     parser.add_argument('-a', "--addresses", required=False, type=str,
-                        help=("Direct TCP portconnections as comma-separated "
-                              "list: 'tcp://127.0.0.1:12345,tcp://123.456.789.0:9013'"))
+                        help=("Add direct TCP port connection.  Can be used several times: "
+                              "'-a tcp://127.0.0.1:12345 -a tcp://123.456.789.0:9013'"),
+                        action="append")
 
     return parser.parse_args()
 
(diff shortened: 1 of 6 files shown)


pnuu:
Needs more testing, currently get:
```python
trollflow2/launcher.py", line 111, in run
    msg = listener.output_queue.get(True, 5)
AttributeError: 'NoneType' object has no attribute 'get'```

**Comment by pnuu** [comment]
Nope, this was caused by the recent(?) addition of passing `topics` as a mandatory command line option.

**Comment by pnuu** [comment]
With a fresh look, the command line arguments work as expected. It's just that the product list has evolved since my previous update and thus doesn't work.

File: bin/satpy_launcher.py [hunk 1]
```diff
@@ -44,11 +45,25 @@
     parser.add_argument("--log-config", '-c',
                         help="Log config file (yaml) to use",
                         type=str, required=False)
+    parser.add_argument('-n', "--nameserver", required=False, type=str,
+                        help="Nameserver to connect to")
+    parser.add_argument('-a', "--addresses", required=False, type=str,
+                        help=("Direct TCP portconnections as comma-separated "
+                              "list: 'tcp://127.0.0.1:12345,tcp://123.456.789.0:9013'"))
```
mraspaud: Could we add `action='append'` on this one instead?

bin/satpy_launcher.py
```diff
@@ -44,11 +45,25 @@
     parser.add_argument("--log-config", '-c',
                         help="Log config file (yaml) to use",
                         type=str, required=False)
+    parser.add_argument('-n', "--nameserver", required=False, type=str,
+                        help="Nameserver to connect to")
```
mraspaud: Maybe a sensible default would be `localhost`

trollflow2/plugins/__init__.py
```diff
@@ -240,11 +240,12 @@ class FilePublisher(object):
     """Publisher for generated files."""
 
     # todo add support for custom port and nameserver
```
mraspaud: This comment should be removed :)

mraspaud:
```suggestion
    # todo add support for custom port and nameserver
```

pnuu:
TODO: The `FilePublisher` kwargs aren't passed from the config to the object. Will investigate more next week.

PR #60, opened 2019-10-14
Check the area from correct level of the product list

pnuu:
This PR fixes the usage of `coverage_by_collection_area` option. The current version wasn't updated when `areas` level was added to the product list.
 - [x] Tests passed 


pnuu:
This is tested (https://github.com/pnuu/trollflow2/blob/bugfix-collection-area-id/trollflow2/tests/test_trollflow2.py#L817), although perhaps not in the best way. The tests passed earlier because the "missing" area ID wasn't on the level where the code tried to find it, and raised the correct `AbortProcessing` exception. I'll add another test that tests also the positive case.

PR #61, opened 2019-11-11
Ensure composites are generated for scenes without resampling

pnuu:
This PR ensures that composites are generated also for `null` areas, i.e. for areas which are saved as-is without resampling.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -123,6 +123,10 @@ def resample(job):
                 job['resampled_scenes'][area] = scn.resample(scn.max_area(),
                                                              **area_conf)
             else:
+                # The composites need to be created for the saving to work
+                if not set(scn.wishlist).issuperset(scn.wishlist):
+                    LOG.debug("Generating composites for 'null' area.")
+                    scn.load(scn.wishlist, generate=True)
                 job['resampled_scenes'][area] = scn
         else:
             LOG.debug("area: %s, area_conf: %s", area, str(area_conf))
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -125,7 +125,7 @@ def resample(job):
             else:
                 # The composites need to be created for the saving to work
                 if not set(scn.datasets.keys()).issuperset(scn.wishlist):
-                    LOG.debug("Generating composites for 'null' area.")
+                    LOG.debug("Generating composites for 'null' area (satellite projection).")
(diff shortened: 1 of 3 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -123,6 +123,10 @@ def resample(job):
                 job['resampled_scenes'][area] = scn.resample(scn.max_area(),
                                                              **area_conf)
             else:
+                # The composites need to be created for the saving to work
+                if not set(scn.datasets.keys()).issuperset(scn.wishlist):
+                    LOG.debug("Generating composites for 'null' area.")
```
mraspaud:
```suggestion
                    LOG.debug("Generating composites for 'null' area (satellite projection).")
```

trollflow2/plugins/__init__.py
```diff
@@ -123,6 +123,10 @@ def resample(job):
...
+                    scn.load(scn.wishlist, generate=True)
```
mraspaud: You are not passing `area_conf`, could this be a problem ? Maybe just pass it with a modified `generate` value ?

pnuu: Are there parameters in `area_conf` that would affect loading for the full disk (or swath) area? In `load_composites()` the only additional kwarg passed to `scn.load()` is `resolution`, so maybe that is needed?

mraspaud: Yes, and other read-specific keyword, eg masking space-pixels for H8...

pnuu: So those other read specific kwargs are also missing from the `load_composites()` call to `scn.load()`? I'll add the resolution kwarg here, the others need another PR (not sure what are really needed in which case).

mraspaud: ah, sorry, I see now that `area_conf` is passed to the resampling

mraspaud: so we probably just need to pickup the parameters from `load_composites` here instead.

pnuu: Oh, but all the relevant stuff (resolution etc.) are already present in `scn.wishlist` which contains a list of `DatasetID`s.

mraspaud: ok then. For reference, that's what is done in `load_composites`: 
```python
scn.load(composites, resolution=resolution, generate=generate)
```

PR #63, opened 2019-11-12
Fix publisher stopping after first processed area

pnuu:
This PR fixes the publisher, which was stopped after the first processed areas. This caused the next area to have `None` as the publisher and a crash.

 - [x] Tests added  
 - [x] Tests passed 


pnuu:
Tested this with live data, and everything seems to work as expected.

PR #64, opened 2019-11-13
Use context manager to create publisher

pnuu:
This PR changes `FilePublisher` so that it starts a publisher when called. This is done to be sure the publisher will be stopped after usage, and will not be recreated afterwards by mistake.

 - [ ] Tests added  
 - [ ] Tests passed 


mraspaud:
Superseeded by #65

PR #66, opened 2019-11-15
Fix formats dictionary being the same object

pnuu:
This PR fixes the `formats` dictionaries being the same object. This caused all the filenames being the same when a minimal product list without explicitly written `formats` section were used, see below.

```yaml
  formats:
    - format: tif
      writer: geotiff
  areas:
      euro4:
        areaname: euro4
        products:
          overview:
            productname: overview
          airmass:
            productname: airmass
```

 - [x] Tests added  
 - [x] Tests passed 


pnuu:
So far this only includes a unittest that highlights the problem. Forgot to select `draft` PR before creating this.

pnuu:
Ok, found a fix!

pnuu:
Confirmed to work also on real life operations.

PR #67, opened 2019-11-21
Use area definition names to check sunlight coverage

pnuu:
This PR changes the behaviour of `check_sunlight_coverage()` so that it will use the name of area definitions instead of individual composite/channel areas read from resampled scene attributes. This way the filtering can be done before loading and resampling. If the data are not to be projected (area defined as `null` in the config), the area is read from the dataset attributes and the plugin needs to be after the data is loaded with `load_composites`.

Closes https://github.com/pytroll/pyresample/issues/228

 - [x] Tests adjusted and added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -521,8 +521,6 @@ def check_sunlight_coverage(job):
             area_def = get_area_def(area)
         except AreaNotFound:
             area_def = None
-        coverage = {True: None, False: None}
-        overpass = None
         for product in products:
             prod_path = "/product_list/areas/%s/products/%s" % (area, product)
             config = get_config_value(product_list, prod_path, "sunlight_coverage")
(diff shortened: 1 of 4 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -573,6 +585,29 @@ def _get_sunlight_coverage(area_def, start_time, overpass=None):
...
+        else:
+            scn = job['scene']
+
+        if isinstance(product, tuple):
+            prod = scn[product[0]]
+        else:
+            prod = scn[product]
+    except KeyError:
+        try:
+            prod = scn[list(scn.keys())[0]]
```
mraspaud: What happens if some of the datasets aren't complete (like 3a or 3b on avhrr ?)

pnuu: I guess then the user could move the plugin to be run after resampling.

pnuu:
I've now tested this wih live data using an named area definition, and directly saving to satellite projection (area set to `null` in yaml).

pnuu:
The coverage calculation doesn't work for projections where the area has space in it, like the full disk `geos` projection for MSG/SEVIRI, Himawari-8/AHI and GOES/ABI. I'll see if I can find a workaround for this case.

pnuu:
Ok, found a solution and it seems to work. I'll check the status in the morning.

pnuu:
Yes, works nicely.

pnuu:
The ZMQ and crash handling commits should have been in a different branch -> reverted.

PR #70, opened 2019-12-03
Use get_geostationary_bounding_box from pyresample instead of satpy

pnuu:
With this PR the imports for GEO bounding box computations are done from Pyresample instead of Satpy. The version there can handle also projections with `ellps` instead of `a/b/h` in the projection definition.

 - [ ] Tests passed 

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -39,7 +39,7 @@ from satpy import Scene
 from satpy.dataset import DatasetID
 from satpy.resample import get_area_def
 from satpy.writers import compute_writer_results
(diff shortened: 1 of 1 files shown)


djhoese:
Should the function in satpy be removed?

pnuu:
Yes. Or at least deprecated and imported instead from Pyresample.

pnuu:
The general functionality has been in Pyresample for 2+ years. A fix for projections using `ellps` instead of `a/b/h` is 2 months old, and doesn't seem to be released yet. That functionality isn't in Satpy either, so doesn't matter here. Merging.

PR #71, opened 2019-12-04
Implement call to native resampler

pnuu:
This PR adds a possibility to call `native` resampler for unprojected (area `None`) scene.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -119,12 +119,15 @@ def resample(job):
             maxarea = get_config_value(product_list,
                                        '/product_list/areas/' + str(area),
                                        'use_max_area')
+            native = conf.get('resampler') == 'native'
             if minarea is True:
                 job['resampled_scenes'][area] = scn.resample(scn.min_area(),
                                                              **area_conf)
             elif maxarea is True:
                 job['resampled_scenes'][area] = scn.resample(scn.max_area(),
                                                              **area_conf)
+            elif native:
+                job['resampled_scenes'][area] = scn.resample(resampler='native')
             else:
                 # The composites need to be created for the saving to work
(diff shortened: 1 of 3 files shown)


pnuu:
Tested with live GOES-16 data from Eumetcast. Works. Without this the composites using downsampled data were not created due to mismatch between areas.

trollflow2/plugins/__init__.py
```diff
@@ -119,12 +119,15 @@ def resample(job):
             maxarea = get_config_value(product_list,
                                        '/product_list/areas/' + str(area),
                                        'use_max_area')
+            native = conf.get('resampler') == 'native'
```
mraspaud: could we make this generic to support other resamplers like ewa or gradient ? and pass their kwargs too ?

pnuu: The basic calls to other resampler are already handled in the non-`None` branch (`else:`) of the resampler. Although, the non-kdtree based resample kwargs need to be handled somehow, currently there's only handling for `radius_of_influence`, `cache_dir` etc. I think that is best done when we get the gradient resampler merged in pyresample and satpy.

mraspaud: ok

djhoese:
What about handling the non-max resolution area? In Polar2Grid I ended up doing a special string "MIN" and "MAX" to mean `scn.min_area()` and `scn.max_area()`.

pnuu:
Both min and max areas are already handled: https://github.com/pnuu/trollflow2/blob/f810ba124e63c22ab959b959c8bcb35cba4800b6/trollflow2/plugins/__init__.py#L123-L128

PR #72, opened 2020-01-28
Change 'import dpath' to 'import dpath.util'

hundahl:
Importing has changed with dpath-2.0.1

### DIFF
--- a/trollflow2/dict_tools.py
+++ b/trollflow2/dict_tools.py
@@ -20,7 +20,7 @@
 # are not necessary
 """Tools for product list operations."""
 
-import dpath
+import dpath.util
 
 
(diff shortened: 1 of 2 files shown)


mraspaud:
https://github.com/akesterson/dpath-python/issues/119

mraspaud:
This seems to be a bug in dpath >= 2.0, so I'm going to wait for an answer from the dpath devs before taking action on this one.

mraspaud:
For now, just install trollflow2 with dpath < 2.0 works

pnuu:
I've forced the dpath version to < 1.5 in #73 which is now merged. I'll close this, and we can do further updates when the dpath issue @mraspaud reported and linked has been handled.

PR #73, opened 2020-02-13
Add a pluging to check message metadata

pnuu:
This PR adds a plugin that can be used to check the sensor given in the message metadata.

This is needed when handling messages coming from `aapp_dr_runner` which basically needs to use the same topics for every sensor.

 - [x] Tests added  
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -425,7 +425,7 @@ def check_sensor(job):
     """Check if the sensor is valid.  If not, discard the scene."""
     mda = job['input_mda']
     product_list = job['product_list']
-    conf = get_config_value(product_list, '/product_list', 'processed_platforms')
+    conf = get_config_value(product_list, '/product_list', 'processed_sensors')
     if conf is None:
         return
     sensor = mda['sensor']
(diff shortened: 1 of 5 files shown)


mraspaud:
I think the dpath should be <2.0, right ?

trollflow2/plugins/__init__.py
```diff
@@ -421,6 +421,19 @@ def check_platform(job):
...
+    """Check if the sensor is valid.  If not, discard the scene."""
+    mda = job['input_mda']
+    product_list = job['product_list']
+    conf = get_config_value(product_list, '/product_list', 'processed_sensors')
+    if conf is None:
+        return
+    sensor = mda['sensor']
+    if sensor not in conf:
+        raise AbortProcessing(
+            "'%s' not in list of allowed sensors" % str(sensor))
```
mraspaud: Should we have a generic metadata checker instead ?
so in the product list:
```yaml
...
processed_metadata:
  sensors: ['viirs', 'modis']
  platform: ['EOS-Terra', 'NOAA-20']
```

pnuu: Yes, I'll do that.

pnuu: I refactorized these two to a single generic `check_metadata` plugin. I'll add the relevant documentation to https://github.com/pytroll/trollflow2/pull/75

PR #75, opened 2020-02-17
Document plugins

### LINKED ISSUE #74: Add documentation on configuration options
## Feature Request

All the plugins and their configuration options should be documented. Not only as a example config, but all the individual plugins should have proper documentation:

- what are they used for
- in which place(s) they should or can be in the list of workers
- all the config options they have

Add the plugin documentation to RTD.

pnuu:
This PR adds documentation on the various Trollflow2 plugins.

 - [x] Closes #74  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -243,7 +243,10 @@ Example
     use_extern_calib: false
     fname_pattern: &fname
       "{platform_name}_{start_time:%Y%m%d_%H%M}_{areaname}_{productname}.{format}"
-    publish_topic: /raster/2A/avhrr
+    subscribe_topics:
+      - /incoming/topic1
+      - /incoming/topic2
+    publish_topic: /raster/2A/avhrr/{areaname}/{productname}
     reader: avhrr_l1b_aapp
     mask_area: True
     delay_composites: True
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -39,6 +39,9 @@ end of each run (one scene).
 An example of such a callable class used in trollflow2 is the
 `class:FilePublisher`.
 
+Most of the configuration are done within the product list.  The
+notable exception are the initialization options for the class-based
+plugins.
 
 Main plugins
 ++++++++++++
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -132,6 +132,9 @@ The available options, with the default values, are:
  - `nameservers: null` - a list of nameservers to connect to.  Default
    to the nameserver on localhost.
 
+These options are given in the list of workers as a dictionary
+argument to the `FilePublisher` class.
+
 Auxiliary plugins
 +++++++++++++++++
 
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -114,6 +114,7 @@ are several options:
    - `format: tif`
    - `writer: geotiff`
    - `fname_pattern: '{platform_name}_{start_time:%Y%m%d_%H%M}_{areaname}_{productname}.{format}'`
+ - `use_tmp_file: bool` - First save the data to a temporary filename and then rename to the final version.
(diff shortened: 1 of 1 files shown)


doc/source/index.rst
```diff
@@ -40,17 +40,202 @@ An example of such a callable class used in trollflow2 is the
 `class:FilePublisher`.
 
 
-Available plugins
-+++++++++++++++++
+Main plugins
+++++++++++++
 
 .. currentmodule:: trollflow2.plugins
```
mraspaud: My hope was that this would include the docstrings of the different plugins here. Obviously this isn't working according to the latest version of the docs published on rtd.

So either we make this work or we remove it and use your text below. Or we use what you wrote as docstrings :)

doc/source/index.rst
```diff
@@ -40,17 +40,202 @@ An example of such a callable class used in trollflow2 is the
...
 
-The `check_sunlight_coverage` plugin
-************************************
+Scene creation
+**************
+
+The `create_scene` plugin creates a scene object.  The configuration
+options are:
```
mraspaud: Should we make it clear that the configuration options are to be specified in the product list, and not in the `workers` section ?

doc/source/index.rst
```diff
@@ -40,17 +40,202 @@ An example of such a callable class used in trollflow2 is the
...
+ - `radius_of_influence: null`
+ - `reduce_data: True`
+ - `cache_dir: None`
+ - `mask_area: False`
+ - `epsilon: 0.0`
+
+For dynamic area definitions, which are not named, one can define the area as `null` and define one of
+
+ - `use_min_area: True` or
+ - `use_max_area: True`
```
mraspaud: I wouldn't call this a dynamic area definition, it's just the satellite projection with pixels aggregated or repeated to make the mulitple resolutions match.

doc/source/index.rst
```diff
@@ -40,17 +40,202 @@ An example of such a callable class used in trollflow2 is the
...
+If there are further software that need to receive Posttroll messages
+of the saved datasets, `FilePublisher` can be used.  This is typically
+the last plugin in the chain, right after `save_datasets`.
+
+The available options, with the default values, are:
+
+ - `port: 0` - publish messages in this port.  By default a random
+   port is selected.
+ - `nameservers: null` - a list of nameservers to connect to.  Default
+   to the nameserver on localhost.
```
mraspaud: I nameservers is left out (default), multicasting is used. (not localhost)

doc/source/index.rst
```diff
@@ -40,17 +40,202 @@ An example of such a callable class used in trollflow2 is the
...
+
+Sensor name check
+*****************
+
+The `check_sensor` plugin can be used to check that the sensor name
+in the incoming message is in the configured list of sensors.  The
+best place for this plugin is before any data handling is done.
+
+Options:
+ - `processed_sensors: null` - A list of allowed sensor names.  By default (`null`) all sensors are processed.
```
mraspaud: Just a thought: could these plugins be replaced by a metadata checker ? (Nothing that should be done in this PR obviously)

doc/source/index.rst
pnuu: Yes, good point.

doc/source/index.rst
pnuu: We discussed this misunderstanding in Slack, and what the `min_area` and `max_area` really mean, and the confusion the adopted naming for them cause (`area` when they really are a matter of `resolution`). I'll rephrase and point to Satpy/Pyresample docs.

doc/source/index.rst
pnuu: Yes, the same crossed my mind.

PR #79, opened 2020-04-30
Add the info parameter to `format`

mraspaud:
This PR adds an `info` field in the `format` sections. This field is not getting out of the product list in any way, but the reason it is not a comment is because we want to be able to manipulate it with a yaml library in the future (for example to be able to modify the product list with a UI).

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -270,6 +270,10 @@ Product list
  - ``productname``, ``areaname`` are the names to use for product and area
    in the filename. If not provided, they default to the actual product
    and area names.
+ - Each ``format`` in the ``formats`` section can have an ``info`` item to store
+   information that will be strictly kept in the product list (ie not saved to
+   file, not published). This can be useful for example for storing information
+   about which customer is receiving this file in the end.
 
 Example
 *******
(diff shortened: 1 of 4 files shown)


pnuu:
I think this should be defined somewhere else in the YAML. I know this is the logical place because how the YAML is handled. From the user and configuration point of view the `info` isn't in anyway connected to the `formats` section which defines the settings for output file formats. _If_ the `info` would be passed to the writer, then I'd agree, but not in this use case.

mraspaud:
@pnuu My idea is that the format section actually is a file section. And the info I want to add is about the files themselves (ie hwo is consuming them)

pnuu:
So we shouid rename the `format` section to `files` (or something more descriptive) then?

mraspaud:
Not sure. At the same time, trollflow2 doesn't support writing multiple files with the same format, so 'file' would be misleading

mraspaud:
I don't think this is relevant anymore, closing.

PR #81, opened 2020-05-05
setting mtime of finished file for detection of newly created products

Weathermann:
update mtime before renaming to real filename for distribution tools that recognize new files based on file age

**modify mtime of created product with touch() from pathlib for dist. tools**

When using `use_tmp_file = True` in trollflow2 yaml products will be created with a tempfile. The duration of the begin of creating the file and finish writing can be a relatively long period of time; sometimes up to 40 minutes (NOAA-20, S-NPP).

DWD uses a file distribution system, which distribute new files to a number of customers. Only new products should be distributed. We recognize new files by the modification time by using the find command (e.g. -mmin 1).
The finished product will have the timestamp of file **creation**.
So with long creation durations the distribution program doesn't recognize such products as new files any more and there will be no distribution to our customers.

The desired behaviour is like the `touch` command in Linux.
After the file is completely written, the file mtime should be updated, to indicate, that it is a new file.

No compatibilty breaks expected.

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -22,6 +22,7 @@
 """Trollflow2 plugins."""
 
 import os
+from pathlib import Path
 from contextlib import contextmanager
 from logging import getLogger
 from tempfile import NamedTemporaryFile
@@ -221,6 +222,7 @@ def renamed_files():
     yield renames
 
     for tmp_name, actual_name in renames.items():
+        Path(tmp_name).touch()    # update mtime before renaming to real filename for dist. tools
         os.rename(tmp_name, actual_name)
 
 


Weathermann:
@mraspaud: Do you think it is really necessary to make the touch() configurable?
Wouldn't it be the expected behavior for a file to get the timestamp when it was completed?

mraspaud:
To know when the file was completed one can check the modification time of the file (mtime). It is interesting to know when the file was created too I think, so for my part I'd like to keep the behaviour as it was.

As for the tests,  it should be put in this file here probably: https://github.com/pytroll/trollflow2/blob/master/trollflow2/tests/test_trollflow2.py#L340
 
Also, could you change the title of this PR to tell a bit more about what this PR implements ?

Sorry for being so annoying on your first PR!

mraspaud:
Any news on this PR @Weathermann ?

mraspaud:
@Weathermann yes, I understand the problem, we would just like to have this `touch` call to be optional :)

gerritholl:
> DWD uses a file distribution system, which distribute new files to a number of customers. Only new products should be distributed. We recognize new files by the modification time by using the find command (e.g. -mmin 1).

Are you sure?  It seems to me that it's our own `copy_afd` script that concludes files are too old.  I don't think the automatic file distributor itself has a mandatory age check.  Can't you just include `file_age` in the `distributions.yaml` file and then the problem is avoided?

PR #83, opened 2020-05-28
Add support to dask.distributed

### LINKED ISSUE #82: Add support for dask.distributed
## Feature Request

Trollflow2 should eventually support `dask.distributed` for parallellism. Both local cluster and external cluster should be supported, but also the current default scheduler should be a possibility.

When suitable configuration is given, use `dask.distributed.Client()` to either connect to a external scheduler, or create a local cluster of workers. If `client` config isn't given, the current behaviour of default scheduler should be used.

The configuration could be something like this:

```yaml
client:
  # If scheduler address is given, use that to access the workers
  scheduler_address: <ip>:8786
  # Otherwise, or if the scheduler can't be connected, local cluster is used with the following params.
  # For the missing arguments the defaults will be used.
  local_cluster:
    nprocs: 4
    nthreads: 1
```

Ideally this shouldn't affect the default usage.

**Additional context**
Some things that might affect the implementation and/or use:

- the client will be recreated for every scene when the new process is started
- only some readers/resamplers/writers can be used with `dask.distributed`

pnuu:
This PR makes it possible to use `dask.distributed` with trollflow2. The example config (below, defined in the trollflow2 config file) uses `dask.distributed.Client` to connect to a scheduler and associated workers. If `dask_distributed` isn't in the configuration, the default scheduler is used. This is hopefully generic enough to meet the idea @djhoese gave in #82 for the user to define their own scheduling/cluster classes.

```yaml
dask_distributed:
  class: !!python/name:dask.distributed.Client
  settings:
    address: localhost:8786
```

 - [x] Closes #82 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/examples/pl.yaml
+++ b/examples/pl.yaml
@@ -1,3 +1,11 @@
+# Use external scheduler.  This example uses dask.distributed.Client,
+#   but any other class can be used as long as it returns a client and
+#   registers itself to dask.  The `settings` are passed as keyword
+#   arguments to the class when instantiated.
+# dask_distributed:
+#   class: !!python/name:dask.distributed.Client
+#   settings:
+#     address: localhost:8786
 
 product_list: &product_list
   output_dir: &output_dir
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -217,6 +217,29 @@ def expand(yml):
     return yml
 
 
+def get_dask_client(config):
+    """Create Dask client if configured."""
+    client = None
+
+    try:
+        client_class = config["dask_distributed"]["class"]
+        client = client_class(**config["dask_distributed"]["settings"])
+        try:
+            if not client.ncores():
+                LOG.error("No workers available, reverting to default scheduler")
+                client.close()
+                client = None
+        except AttributeError:
+            pass
+    except OSError:
+        LOG.error("Scheduler not found, reverting to default scheduler")
+    except KeyError:
+        LOG.info("Distributed processing not configured, "
+                 "using default scheduler")
+
+    return client
+
+
 def process(msg, prod_list, produced_files):
     """Process a message."""
     config = {}
@@ -227,6 +250,9 @@ def process(msg, prod_list, produced_files):
         # Either open() or yaml.load() failed
         LOG.exception("Process crashed, check YAML file.")
         raise
+
+    # Get distributed client
+    client = get_dask_client(config)
 
     try:
         config = expand(config)
@@ -258,6 +284,10 @@ def process(msg, prod_list, produced_files):
             except AttributeError:
                 continue
         del config
+        try:
+            client.close()
+        except AttributeError:
+            pass
         gc.collect()
 
 
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -223,7 +223,8 @@ def get_dask_client(config):
 
     try:
         client_class = config["dask_distributed"]["class"]
-        client = client_class(**config["dask_distributed"]["settings"])
+        settings = config["dask_distributed"].get("settings", {})
+        client = client_class(**settings)
         try:
             if not client.ncores():
                 LOG.error("No workers available, reverting to default scheduler")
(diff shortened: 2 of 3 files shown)


trollflow2/launcher.py
```diff
@@ -217,6 +217,29 @@ def expand(yml):
...
+def get_dask_client(config):
+    """Create Dask client if configured."""
+    client = None
+
+    try:
+        client_class = config["dask_distributed"]["class"]
+        client = client_class(**config["dask_distributed"]["settings"])
+        try:
+            if not client.ncores():
+                LOG.error("No workers available, reverting to default scheduler")
```
mraspaud, line 229: What about just warning ? Maybe the workers are started after ?

trollflow2/launcher.py
```diff
@@ -217,6 +217,29 @@ def expand(yml):
     return yml
 
 
+def get_dask_client(config):
+    """Create Dask client if configured."""
+    client = None
+
+    try:
+        client_class = config["dask_distributed"]["class"]
+        client = client_class(**config["dask_distributed"]["settings"])
```
mraspaud, line 226: How about leaving this empty if no settings are provided ? So the default behaviour will be used (the scheduler and worker is created automatically, good for testing)

pnuu: When this setting isn't in the config we'll use the default scheduler: no local cluster, no clients. That is, the current behaviour, which is often the most efficient in any case.

mraspaud: ok, I mean that if the scheduler address isn't specified, the distributed scheduler would still be used but with default parameters. If the `dask_distributed` setting, or `class` isn't defined, we fallback to the default scheduler.

pnuu: Ah, ok. That's what happens here. If I define the config like this:
```yaml
dask_distributed:
  class: !!python/name:dask.distributed.Client
  settings: {}
```
a local cluster is created with default parameters. In the settings the user can also define any keyword argument the `Client` class supports for creating a local cluster.

pnuu: Actually, I'll make the `settings` optional to make the config simpler.

examples/pl.yaml
```diff
@@ -1,3 +1,11 @@
+# Use external scheduler.  This example uses dask.distributed.Client,
+#   but any other class can be used as long as it returns a client and
+#   registers itself to dask.  The `settings` are passed as keyword
+#   arguments to the class when instantiated.
+# dask_distributed:
+#   class: !!python/name:dask.distributed.Client
```
djhoese, line 6: Are there multiple Client classes? I know there are multiple Cluster classes for the different types of clusters (slurm, kubernetes, etc). The Cluster object is given to the Client object when you create it.

pnuu: There is only one `Client` class in `dask.distributed`, which can be given the address of the scheduler it uses. If the address is not given, it'll create a local cluster based on the other keyword arguments. Looking at https://docs.dask.org/en/latest/setup/single-distributed.html the hand-created clusters can be given as the address kwarg to the client, and the `Client` docstring says:
```
address: string, or Cluster
    This can be the address of a ``Scheduler`` server like a string
    ``'127.0.0.1:8786'`` or a cluster object like ``LocalCluster()``
```

I guess if non-address non-local clusters are wanted, they could be defined via YAML magic in the `settings` dictionary.

djhoese: Yeah that might work. It might we hard to specify all of the `__init__` arguments for creating a cluster:

```
from dask_yarn import YarnCluster
from dask.distributed import Client

# Create a cluster where each worker has two cores and eight GiB of memory
cluster = YarnCluster(environment='environment.tar.gz',
                      worker_vcores=2,
                      worker_memory="8GiB")
# Scale out to ten such workers
cluster.scale(10)

# Connect to the cluster
client = Client(cluster)
```

But I see that some clusters can be configured by other YAML files (hadoop example shows that) so maybe that could be used too.

Anyway, I'm not sure there is a need for the class parameter here in this YAML then, right? At least not for the Client.

trollflow2/launcher.py
pnuu: They could be started later, but we aren't waiting for them. But I'll make it a warning in any case.

PR #58, opened 2019-10-04
Fix versioning using setuptools_scm and leave versioneer.py

adybbroe:
Replace versioneer.py using setuptools_scm instead. versioneer is not maintained.

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Tests passed 
 - [ ] Passes ``flake8 trollflow2`` 
 - [ ] Fully documented 
 - [x] Add your name to `AUTHORS.md` if not there already

PR #85, opened 2020-07-16
Fix broken scm versioning

adybbroe:
This PR tries to fix an unfinished PR #58 to swicth to use setuptools_scm for git tag versioning

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented 


setup.py
```diff
@@ -56,14 +56,16 @@
                    "Programming Language :: Python",
                    "Topic :: Scientific/Engineering"],
       url="https://github.com/pytroll/trollflow2",
-      packages=['trollflow2',
-                'trollflow2.plugins',
-                ],
+      packages=find_packages(),
+      # packages=['trollflow2',
+      #          'trollflow2.plugins',
+      #          ],
```
mraspaud, line 62: why leave these comments ?

adybbroe: I was not sure find_packages would find the plugins, that is why. Don't know what will need this?
Can we test/verify before I remove the old lines?

PR #89, opened 2020-09-24
Fix null resampling with newer versions of satpy

### LINKED ISSUE #88: Failure when saving in satellite-native projection
**Describe the bug**
Processing fails when using satellite-native projection, that is `null` as projection in `trollflow.yaml`.

**To Reproduce**
Use `null` as area, like in the following stripped-down non-working product list:

```yaml
product_list:
  areas:
    null:
      areaname: geos
      products:
        airmass:
          productname: airmass

```

**Expected behavior**
The images should be saved without resampling.

**Actual results**
Process crashes with

```python
  File "/home/users/satman/current/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/users/satman/current/lib/python3.8/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/users/satman/current/lib/python3.8/site-packages/trollflow2/launcher.py", line 268, in process
    cwrk.pop('fun')(job, **cwrk)
  File "/home/users/satman/current/lib/python3.8/site-packages/trollflow2/plugins/__init__.py", line 136, in resample
    if not set(scn.datasets.keys()).issuperset(scn.wishlist):
AttributeError: 'Scene' object has no attribute 'datasets'
```

**Environment Info:**
 - Trollflow2 Version: 0.7.2.dev19+g4158921 (latest master)
 - Satpy Version: 0.23.0

mraspaud:
Satpy 0.23 privatized the `.datasets` attribute of the scene, and thus broke some parts of trollflow2 that used it. This PR fixes it.

 - [x] Closes #88 
 - [x] Tests added  
 - [x] Tests passed

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -133,7 +133,7 @@ def resample(job):
                 job['resampled_scenes'][area] = scn.resample(resampler='native')
             else:
                 # The composites need to be created for the saving to work
-                if not set(scn.datasets.keys()).issuperset(scn.wishlist):
+                if not set(scn.keys()).issuperset(scn.wishlist):
                     LOG.debug("Generating composites for 'null' area (satellite projection).")
                     scn.load(scn.wishlist, generate=True)
                 job['resampled_scenes'][area] = scn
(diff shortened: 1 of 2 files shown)


pnuu (approved):
Works! Thanks for fixing this :smile:

PR #97, opened 2021-02-03
Adapt trollflow2 for removal of ppp_config_dir

gerritholl:
Adapt the trollflow2 `create_scene` plugin for the removal of
ppp_config_dir.  Fixes #96.

 - [x] Closes #96 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Passes ``flake8 trollflow2`` 
 - [x] Add your name to `AUTHORS.md` if not there already

PR #106, opened 2021-03-16
Remove the ppp_config_dir scene creation parameter for satpy>=0.26.0 compatibility

mraspaud:
As of satpy 0.26.0, the scene creation doesn't accept the `ppp_config_dir` parameter anymore.
This PR removes it, since the prefered usage is to path `SATPY_CONFIG_DIR` as an environment variable now.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented


mraspaud:
duplicate of #97

djhoese:
FYI It is `SATPY_CONFIG_PATH` not `DIR`.

PR #109, opened 2021-03-30
Check that products exist before publishing

### LINKED ISSUE #77: Sometimes weird products is send to plugins.FilePublisher
I noticed that the plugins.FilePublisher once in a while is publishing the AAPP l1b file (with a weird path), fx:

`"uri": "/opt/aapp8/AAPP/orbelems/hrpt_metop03_20200313_1231_06996.l1b",
"uid": "hrpt_metop03_20200313_1231_06996.l1b"
"product": "ir_037"
"productname": "pol_ir3.7"
"format": "tif"`

The product pol_ir3.7 is not been produced and is been skipped in plugings.save_dataset. Could it be that skipped products are not been removed from the final product list send to plugins.FilePublisher ? It's also seen for other product, fx. cloudtop

Also, it looks hard to filter out in a trollmoves.dispatcher config file, since it has all the right tags

mraspaud:
This PR fixes the FilePublisher by looking if a product exists in the resampled scene prior to publishing.

 - [x] Closes #77 
 - [x] Tests added  
 - [x] Tests passed

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -337,7 +337,7 @@ class FilePublisher(object):
         mda.pop('collection', None)
         for fmat, fmat_config in plist_iter(job['product_list']['product_list'], mda):
             if fmat['product'] not in job['resampled_scenes'].get(fmat['area'], []):
-                LOG.debug('Skipped product not published %s.', str(fmat))
+                LOG.debug('Not publishing missing product %s.', str(fmat))
                 continue
             try:
                 topic, file_mda = self.create_message(fmat, mda)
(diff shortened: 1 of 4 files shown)


pnuu (approved):
LGTM

trollflow2/plugins/__init__.py
```diff
@@ -336,6 +336,9 @@ def __call__(self, job):
         mda.pop('dataset', None)
         mda.pop('collection', None)
         for fmat, fmat_config in plist_iter(job['product_list']['product_list'], mda):
+            if fmat['product'] not in job['resampled_scenes'].get(fmat['area'], []):
+                LOG.debug('Skipped product not published %s.', str(fmat))
```
mraspaud, line 340:
```suggestion
                LOG.debug('Not publishing missing product %s.', str(fmat))
```

PR #110, opened 2021-04-01
For unstable build, also install pyresample.

gerritholl:
For the unstable build in which satpy is installed without dependencies,
also install pyresample, or importing satpy may fail under some
circumstances.

 - [ ] Tests added  
 - [x] Tests passed 


mraspaud (approved):
LGTM

PR #112, opened 2021-04-23
Skip products when data contain too many fill values

### LINKED ISSUE #111: Check channel-dependent validity
## Feature Request

We're processing AVHRR data.  AVHRR may switch between channels 3A and 3B in the middle of a swath.  This means both channels 3A and 3B and their composites can be loaded.  After resampling to the area of interest, up to 100% of the resampled scene may be invalid (NaN) for either of those channels and composites using them.  Pytroll will write images containing no valid data.  In our client software (NinJo), these images will be shown as available, but when loading, nothing will be shown (because everything is fill value), and the user will think something is wrong.  I'd like that trollflow2 can skip products that are mostly or entirely fill value.

I'd like to add a plugin `check_valid` or something like that, that verifies that at least x% (configurable) of a product (channel or composite) is valid (either by `.notnull()` or by `.notnull()/covers()`).  If not, this product is removed (similar to sunlight coverage checks).

Fully backward compatible.

**Additional context**

I could remove files that are suspiciously small, but this kind of check should be done before, not after, saving datasets.

gerritholl:
Plugin to check validity of channel data before continuing processing.  This will help us when AVHRR switches 3A/3B so we don't generate empty channels and products.

 - [ ] Closes #111 
 - [x] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -675,3 +675,7 @@ def _get_plugin_conf(product_list, path, defaults):
         conf[key] = get_config_value(product_list, path, key,
                                      default=defaults.get(key))
     return conf
+
+
+def check_valid(job):
+    raise NotImplementedError()
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -1516,5 +1516,28 @@ class TestFilePublisher(TestCase):
         nb_.stop.assert_called_once()
 
 
+def test_valid_filter():
+    """Test filter for minimum fraction of valid data."""
+    from trollflow2.launcher import yaml, UnsafeLoader
+    from trollflow2.plugins import check_valid
+    product_list = yaml.load(yaml_test3, Loader=UnsafeLoader)
+    scene = mock.MagicMock()
+    resampled_scene = mock.MagicMock()
+    job = {}
+    job['scene'] = scene
+    job['product_list'] = product_list.copy()
+    job['input_mda'] = input_mda.copy()
+    job['resampled_scenes'] = {"euron1": resampled_scene}
+    prods = job['product_list']['product_list']['areas']['euron1']['products']
+    prods['green_snow']["min_valid"] = 10
+    for p in ("IR016", "IR037", "overview"):
+        prods[p] = {"min_valid": 10}
+    check_valid(job)
+    assert "green_snow" not in resampled_scene
+    assert "IR016" not in resampled_scene
+    assert "IR037" in resampled_scene
+    assert "overview" in resampled_scene
+
+
 if __name__ == '__main__':
     unittest.main()


trollflow2/plugins/__init__.py
```diff
@@ -675,3 +675,72 @@ def _get_plugin_conf(product_list, path, defaults):
...
+                start_time = prod.attrs["start_time"]
+                end_time = prod.attrs["end_time"]
+                sensor = prod.attrs["sensor"]
+                if area_name not in exp_cov:
+                    # get_scene_coverage uses %, convert to fraction
+                    exp_cov[area_name] = get_scene_coverage(
+                        platform_name, start_time, end_time, sensor, area_name)/100
+                valid = job["resampled_scenes"][area_name][prod_name].notnull()
+                exp_valid = exp_cov[area_name]
+                actual_valid = float(valid.sum()/valid.size)
```
gerritholl, line 729: This line is slow, because it computes the data.  I warn for this in the documentation.  However, it appears the computation is not persistent: `save_datasets()` is as slow as before, and overall processing is therefore much slower than before.  How can I make sure dask remembers this computation?

gerritholl:
On hold considering the question on the optimal approach here.  As implemented as present, data are calculated twice, which almost doubles the total time to generate products.  This is not acceptable.

Alternatives:

- Find a way similar to what I'm doing now (check the data), but make sure the products are only calculated once, not twice.  This could be done by deferring the checks to just before `save_datasets`.  Advantage: broadly applicable (validity checks covers not only AVHRR 3A/3B, but any bad/flagged data), and I've already partly implemented it.  Disadvantage: channel data that aren't there are still resampled, which is a waste of time.
- Use per-channel start/end times in the to calculate per-channel area coverage.  This would require changing all the AVHRR readers (or at least the ones we're using) to adapt per-channel start/end times and changing the `covers()` plugin to calculate coverage per group of products sharing a start/end time.

The second alternative seems cleaner, but also more work.  The first alternative would also cover cases with excessive bad data other than absent channels, but I don't know if that use case is needed.

gerritholl:
I think I need a combination of both.  Although https://github.com/pytroll/satpy/pull/1653 works now and could be used in a new or adapted trollflow2 plugin, there are still remaining issues.  The start and end times are still valid only for the original (set of) files read; they are not updated after resampling, and updating them after resampling would be somewhat involved.  Secondly, some data files (such as `hrpt_noaa19_20210423_1411_62891.l1b`) lie about the presence of 3A/3B, so even with this satpy PR, the data are claimed to be there all the time (according to the 3a/3b bit), but data contents are empty/fill value.

A solution could be a combined approach.  First I do a check how present a certain channel is, using the new updated `start_time` and `end_time` attributes.  Then, after or inside the `save_datasets` plugin, I reject channels with too much invalid data.

gerritholl:
Despite the passing unit tests, this is not yet function.  I don't understand how the unit tests pass, but the new log message reveals in practice "Found 0 unique start/end time pair(s) in scene".  Needs work, next week.

gerritholl:
I've given up on trying to use product-specific `start_time` and `end_time` as being too complicated and an insufficient solution.  Currently, `covers()` can be done before loading any datasets.  A product-specific `covers()` would need to be done after loading datasets, because it needs the `start_time` and `end_time` for each dataset.  That means the plugin should be a separate function in additional to `covers()`.  But the `start_time` and `end_time` are not updated when resampling.  Some products may not have a `start_time` and `end_time`.  And even if I could get it to work, it wouldn't replace the issue with filtering, becauset he data source isn't always truthful about what channel (3A or 3B) is active.

So back to the content filtering solution, now in `check_valid()` I'm replacing the dataset as stored in the product list by the results of its `..persist()` method, which should hopefully prevent double calculation.  Nevertheless, it's still much slower to use `check_valid()` than to not use it: for one example file (20210503135536_NOAA_19.hmf), it takes 01:02 to produce 36 files when skipping `check_valid()` and 02:35 when using `check_valid()`.  I don't understand why.

gerritholl:
With the latest commit, producing 36 files for `20210503135536_NOAA_19.hmf` takes 01:21.  Without `check_valid()` it takes 01:02.  With the old `check_valid()` from earlier commits it takes 02:35.

trollflow2/plugins/__init__.py
gerritholl: No longer relevant; everything is persisted at once in a dedicated function.

gerritholl:
I don't know how to solve the CodeFactor and codecov/project issues.  Any hints?

mraspaud:
One solution might be to start having each plugin in a separate file

gerritholl:
> One solution might be to start having each plugin in a separate file

Good idea (I don't like functional code in `__init__.py` anyway).  Can we postpone that to another PR?

PR #115, opened 2021-05-11
Use queued logging for running in subprocesses

mraspaud:
This PR fixes the log passing from subprocesses to the main process by using a QueueHandler instead of using the predifined handlers passed on through the forking mechanism. This is the recommended way of doing the logging across processes.
The current implementation is suspected to be at the origin of deadlocks causing trollflow2 processing to stop and hang forever, as observed by @gerritholl and myself.

The runner code and tests were also refactored for cleaning purposes.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [ ] Fully documented

### DIFF
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -58,7 +58,6 @@ def main():
     """Launch trollflow2."""
     args = vars(parse_args())
 
-
     log_config = args.pop("log_config", None)
     if log_config is not None:
         with open(log_config) as fd:
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -56,15 +56,18 @@ def parse_args():
 
 def main():
     """Launch trollflow2."""
-    args = parse_args()
-    prod_list = args.product_list
-    test_message = args.test_message
-    topics = args.topic
-    nameserver = args.nameserver
-    addresses = args.addresses
+    args = vars(parse_args())
+    print(args)
 
-    if args.log_config is not None:
-        with open(args.log_config) as fd:
+    # prod_list = args.product_list
+    # test_message = args.test_message
+    # topics = args.topic
+    # nameserver = args.nameserver
+    # addresses = args.addresses
+
+    log_config = args.pop("log_config", None)
+    if log_config is not None:
+        with open(log_config) as fd:
             import yaml
             log_dict = yaml.load(fd.read())
             logging.config.dictConfig(log_dict)
@@ -72,8 +75,11 @@ def main():
         from satpy.utils import debug_on
         debug_on()
 
-    run(prod_list, topics=topics, test_message=test_message,
-        nameserver=nameserver, addresses=addresses)
+    product_list = args.pop("product_list")
+    test_message = args.pop("test_message")
+    connection_parameters = args
+
+    run(product_list, connection_parameters, test_message)
 
 
 if __name__ == "__main__":
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -57,7 +57,6 @@ def parse_args():
 def main():
     """Launch trollflow2."""
     args = vars(parse_args())
-    print(args)
 
     # prod_list = args.product_list
     # test_message = args.test_message
--- a/bin/satpy_launcher.py
+++ b/bin/satpy_launcher.py
@@ -58,11 +58,6 @@ def main():
     """Launch trollflow2."""
     args = vars(parse_args())
 
-    # prod_list = args.product_list
-    # test_message = args.test_message
-    # topics = args.topic
-    # nameserver = args.nameserver
-    # addresses = args.addresses
 
     log_config = args.pop("log_config", None)
     if log_config is not None:
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -30,16 +30,19 @@ memory buildup.
 import ast
 import copy
 import gc
+import logging.handlers
 import os
 import re
 import traceback
 import signal
 from collections import OrderedDict
 from datetime import datetime
-from logging import getLogger
+from logging import getLogger, handlers
+from functools import partial
+from multiprocessing import Queue
 
 import yaml
-from six.moves.queue import Empty
+from queue import Empty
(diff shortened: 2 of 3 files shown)


bin/satpy_launcher.py
```diff
@@ -56,24 +56,23 @@ def parse_args():
...
         from satpy.utils import debug_on
         debug_on()
 
-    run(prod_list, topics=topics, test_message=test_message,
-        nameserver=nameserver, addresses=addresses)
+    product_list = args.pop("product_list")
+    test_message = args.pop("test_message")
+    connection_parameters = args
+
+    run(product_list, connection_parameters, test_message)
```
pnuu, line 75: The return value is not stored. If it is not needed, there shouldn't be explicit `return` in `run()`.

trollflow2/tests/test_launcher.py
```diff
@@ -245,15 +245,48 @@ def setUp(self):
...
+                mock.patch('trollflow2.launcher.process') as process,\
+                mock.patch('trollflow2.launcher.check_results'),\
+                mock.patch('multiprocessing.Process'):
+            generate_messages.side_effect = ['foo', KeyboardInterrupt]
+            prod_list = 'bar'
+            try:
+                run(prod_list)
+            except KeyboardInterrupt:
+                pass
+            assert process.call_count == 0
```
pnuu, line 263:
```suggestion
            process.assert_not_called()
```

trollflow2/tests/test_launcher.py
```diff
@@ -245,15 +245,48 @@ def setUp(self):
...
                 mock.patch('trollflow2.launcher.process') as process,\
+                mock.patch('multiprocessing.Process') as Process:
+            generate_messages.side_effect = ['foo', KeyboardInterrupt]
+            Process.side_effect = Thread
+            prod_list = 'bar'
+            try:
+                run(prod_list)
+            except KeyboardInterrupt:
+                pass
+            assert process.called
```
pnuu, line 281:
```suggestion
            process.assert_called_once()
```

trollflow2/tests/test_launcher.py
```diff
@@ -301,6 +361,74 @@ def test_run_keyboard_interrupt(self):
...
+            Process.side_effect = Thread
+            process.side_effect = fake_process
+            generate_messages.return_value = ['bla']
+
+            try:
+                LOG.addHandler(fake_handler)
+                run(0)
+            finally:
+                LOG.removeHandler(fake_handler)
+            assert len(fake_handler.method_calls) > 0
```
pnuu, line 398:
```suggestion
            assert fake_handler.method_calls
```

trollflow2/launcher.py
```diff
@@ -264,6 +305,23 @@ def get_dask_client(config):
     return client
 
 
+def qlogged_process(msg, prod_list, produced_files, log_queue):
```
pnuu, line 308: Should this have a more descriptive name?
```suggestion
def queue_logged_process(msg, prod_list, produced_files, log_queue):
```

trollflow2/launcher.py
```diff
@@ -115,40 +116,82 @@ def check_results(produced_files, start_time, exitcode):
...
+
+
+def _run_threaded(product_list, test_message):
+    """Run in a thread."""
+    LOG.info("Launching trollflow2 with threads")
+    from threading import Thread
+    from posttroll.message import Message
+    messages = [Message(rawstr=test_message)]
+    target_fun = partial(process, prod_list=product_list)
+    return _run_product_list_on_messages(messages, target_fun, Thread)
```
pnuu, line 159: `_run_product_list_on_message()` doesn't return anything, so this shouldn't either.
```suggestion
    _run_product_list_on_messages(messages, target_fun, Thread)
```

trollflow2/launcher.py
```diff
@@ -115,40 +116,82 @@ def check_results(produced_files, start_time, exitcode):
...
+        topics=topics)
+    return listener
+
+
+def run(product_list, connection_parameters=None, test_message=None):
+    """Spawn one or multiple subprocesses to run the jobs from the product list."""
+    test_message = get_test_message(test_message)
+    if test_message:
+        return _run_threaded(product_list, test_message)
+    return _run_subprocess(product_list, connection_parameters)
```
pnuu, line 149: Neither of these should return anything.

```suggestion
    if test_message:
        _run_threaded(product_list, test_message)
    else:
        _run_subprocess(product_list, connection_parameters)
```

trollflow2/launcher.py
```diff
@@ -115,40 +116,82 @@ def check_results(produced_files, start_time, exitcode):
...
+    target_fun = partial(process, prod_list=product_list)
+    return _run_product_list_on_messages(messages, target_fun, Thread)
+
+
+def _run_subprocess(product_list, connection_parameters=None):
+    """Run in a subprocess, with queued logging."""
+    LOG.info("Launching trollflow2 with subprocesses")
+    from multiprocessing import Process
+    log_queue = Queue()
+    target_fun = partial(qlogged_process, log_queue=log_queue, prod_list=product_list)
```
pnuu, line 167: More descriptive name?

```suggestion
    target_fun = partial(queue_logged_process, log_queue=log_queue, prod_list=product_list)
```

pnuu (approved):
Few suggestions, otherwise LGTM.

I especially like the refactoring that was done, now it's easy to move the test message stuff completely away from the `satpy_launcher.py`. That feature should've originally been in a completely separate script.

bin/satpy_launcher.py
mraspaud: Indeed, thanks!

trollflow2/tests/test_launcher.py
mraspaud: Fixed

mraspaud:
Apparently, `qsize` which we use in check_results isn't implemented on Unix/MacOS, rendering trollflow2 unusable on that platform. @pnuu what do you reckon, should we just drop macos?

mraspaud:
The call to `qsize` was added by @gerritholl but wasn't caught by the unittests as the queue was faked...

PR #124, opened 2021-10-29
Logging fix using global QueueHandler

mraspaud:
It has been observed that trollflow2 v0.11.0 hangs quite often, probably due to a race condition in the logging code.

It is very likely that this bug has been introduced in #115, hence this attempt to make the QueueHandler usage more streamlined, independently of whether trollflow2 runs on threads or subprocesses, and more importantly by setting up the queued log handler and other handlers prior to running the main loop.

EDIT: for completeness, it should be noted that many live tests have been run with v0.10.0 at SMHI and that hanging specifically in the logging part was also happening.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Live test ran several days without crashing
 - [x] Fully documented

### DIFF
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -39,11 +39,10 @@ from contextlib import suppress
 from datetime import datetime
 from functools import partial
 from logging import getLogger
-from multiprocessing import Manager
 from queue import Empty
+from urllib.parse import urlparse
 
 import yaml
-from urllib.parse import urlparse
 from trollflow2.dict_tools import gen_dict_extract, plist_iter
 from trollflow2.logging import setup_queued_logging
 from trollflow2.plugins import AbortProcessing
@@ -62,9 +61,6 @@ except ImportError:
 
 LOG = getLogger(__name__)
 DEFAULT_PRIORITY = 999
-
-MANAGER = Manager()
-LOG_QUEUE = MANAGER.Queue()
 
 
 def tuple_constructor(loader, node):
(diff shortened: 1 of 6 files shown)


trollflow2/logging.py
```diff
@@ -0,0 +1,73 @@
...
+import logging.config
+from logging.handlers import QueueHandler, QueueListener
+from multiprocessing import Queue
+
+DEFAULT_LOG_CONFIG = {'version': 1,
+                      'formatters': {'pytroll': {'format': '[%(levelname)s: %(asctime)s : %(name)s] %(message)s',
+                                                 'datefmt': '%Y-%m-%d %H:%M:%S'}},
+                      'handlers': {'file': {'class': 'logging.StreamHandler',
+                                            'formatter': 'pytroll'}},
+                      'root': {'level': 'DEBUG', 'handlers': ['file']}}
```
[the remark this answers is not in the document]
mraspaud: yep, sorry about that, I think it's fixed now.

trollflow2/launcher.py
```diff
@@ -319,24 +328,27 @@ def get_dask_client(config):
 
 def queue_logged_process(msg, prod_list, produced_files, log_queue):
     """Run `process` with a queued log."""
-    _reset_log_handlers()
-    handler = handlers.QueueHandler(log_queue)
-    LOG.addHandler(handler)
+    setup_queued_logging(log_queue)
+    with suppress(ValueError):
+        signal.signal(signal.SIGUSR1, print_traces)
+        LOG.debug("Use SIGUSR1 on pid {} to check the current tracebacks of this subprocess.".format(os.getpid()))
```
gerritholl, line 334: Does this work only for subprocesses or also for threads?

pnuu, line 334: Should the log message be before the traces are printed? It'll look weird in the logs in this order.

trollflow2/launcher.py
```diff
@@ -319,24 +328,27 @@ def get_dask_client(config):
...
-    try:
-        while LOG.hasHandlers():
-            LOG.removeHandler(LOG.handlers[0])
-    except IndexError:
-        pass
+def print_traces(signum, frame):
+    """Print traces for debugging."""
+    import sys
+    import traceback
+    print(f"Got signal {signum} in {frame}, dumping traces.")
```
pnuu, line 342: Should these prints go to `stderr`?

pnuu (approved):
Thanks for the fix and cleaning the `run()` with the new `Runner` class.

LGTM, just some suggestions.

trollflow2/launcher.py
```diff
@@ -150,66 +150,75 @@ def _create_listener_from_connection_parameters(connection_parameters):
...
+        """Set up the runner."""
+        self.product_list = product_list
+        self.log_queue = log_queue
+        self.connection_parameters = connection_parameters
+        self.test_message = get_test_message(test_message)
+        self.threaded = threaded
 
-    messages = generate_messages(connection_parameters)
+    def run(self):
+        """Spawn one or multiple subprocesses to run the jobs from the product list."""
```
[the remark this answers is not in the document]
mraspaud: Indeed

trollflow2/launcher.py
mraspaud: the `signal.signal` call just sets up the call back to run when the signal happens, so this won't print any traces at that moment.
The reason why the log message happens after is that if that if that function is not called in the main thread of a process, it fails, and thus the log message shouldn't be printed.

trollflow2/launcher.py
mraspaud: ah, yes, good idea

trollflow2/launcher.py
mraspaud: Only for subprocesses unfortunately.

trollflow2/launcher.py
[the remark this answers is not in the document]
mraspaud: Fixed

trollflow2/launcher.py
mraspaud: fixed

PR #126, opened 2021-11-05
Add flag to create temporary file in temporary directory

### LINKED ISSUE #125: Writing ninjo(geo)tiff with use_tmp_file writes temporary filename to metadata
**Describe the bug**

The ninjotiff and ninjogeotiff writers in satpy write the filename to the product headers.  When using trollflow2 with `use_tmp_file = True` (introduced in #41), this becomes the temporary filename.  For example:

```
ninjo_FileName=/data/pytroll/OUTBOXES/generic/tmp22601q9r8v_kj
```

This becomes a problem if NinJo wants to use the filename in any way, such as displaying it to the user.

**To Reproduce**

```python
from satpy import Scene
from glob import glob
from queue import Queue
sc = Scene(
        filenames=seviri_files,
        reader=["seviri_l1b_hrit"])
sc.load(["IR_108"])
ls = sc.resample("eurol")

product_list = {
    "fname_pattern": "name.tif",
    "use_tmp_file": True,
    "areas": {
        "eurol": {
            "products": {
                "IR_108": {
                    "productname": "IR108",
                    "formats": [
                        {"writer": "ninjogeotiff",
                         "ChannelID": 0,
                         "DataType": 0,
                         "PhysicUnit": "no",
                         "PhysicValue": "yes",
                         "SatelliteNameID": 0,
                         }
                    ]
                }
            }
        }
    }
}

job = {}
job['resampled_scenes'] = {"eurol": ls}
job["input_mda"] = {}
job["produced_files"] = Queue()
save_datasets(job)
```

followed by

```bash
gdalinfo name.tif | grep -i filename
```

**Expected behaviour**

ninjo_FileName=name.tif

**Actual results**

ninjo_FileName=/data/gholl/checkouts/protocode/mwe/tmp4679ypf79qdg

**Environment Info:**
 - Trollflow2 Version: 0.11.0
 - Satpy Version: 0.31.0

**Additional context**

Our EBP colleagues want to display in the NinJo client whether the image displayed originates from a GeoTIFF or NinJoTIFF.  This is useful to evaluate the new GeoTIFF format.  The preferred implementation on the NinJo side is via the FileName tag, from which the format can be extracted.  However, the filename as currently written to the headers is wrong.

gerritholl:
Add an option to create the temporary file in a temporary directory, respecting the basename.  This will close #125.  We need this so that the filename passed to the writer and therefore the contents of the filename tag in ninjogeotiff files remains meaningful.

 - [x] Closes #125 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -252,11 +252,16 @@ def renamed_files():
 def save_datasets(job):
     """Save the datasets (and trigger the computation).
 
-    If the `use_tmp_file` option is provided in the product list and is set to
+    If the ``use_tmp_file`` option is provided in the product list and is set to
     True, the file will be first saved to a temporary name before being renamed.
     This is useful when other processes are waiting for the file to be present
     to start their work, but would crash on incomplete files.
 
+    If, in addition, the ``use_tmp_dir`` option is provided in the product
+    list, then the temporary file will be created in this directory without
+    changing the filename.  This is useful for writers which write the filename
+    to the headers, such as the SatPy ninjotiff and ninjogeotiff writers.  The
+    value of ``use_tmp_dir`` has no effect if ``use_tmp_file`` is False.
     """
     scns = job['resampled_scenes']
     objs = []
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -190,14 +190,14 @@ def prepared_filename(fmat, renames):
 
     # tmp filenaming
     use_tmp_file = fmat.get('use_tmp_file', False)
-    use_tmp_dir = fmat.get("use_tmp_dir", False)
+    staging_zone = fmat.get("staging_zone", False)
 
-    if use_tmp_file:
-        if use_tmp_dir:
-            tmp_dir = pathlib.Path(use_tmp_dir)
+    if staging_zone or use_tmp_file:
+        if staging_zone:
+            directory = pathlib.Path(staging_zone)
             of = pathlib.Path(orig_filename)
-            filename = os.fspath(tmp_dir / of.name)
-        else:
+            filename = os.fspath(directory / of.name)
+        if use_tmp_file:
             filename = _get_temp_filename(directory, renames.keys())
         yield filename
         renames[filename] = orig_filename
@@ -257,11 +257,13 @@ def save_datasets(job):
     This is useful when other processes are waiting for the file to be present
     to start their work, but would crash on incomplete files.
 
-    If, in addition, the ``use_tmp_dir`` option is provided in the product
-    list, then the temporary file will be created in this directory without
+    If the ``staging_zone`` option is provided in the product
+    list, then the file will be created in this directory first, without
     changing the filename.  This is useful for writers which write the filename
     to the headers, such as the SatPy ninjotiff and ninjogeotiff writers.  The
-    value of ``use_tmp_dir`` has no effect if ``use_tmp_file`` is False.
+    ``staging_zone`` directory must be on the same filesystem as
+    ``output_dir``.  It is recommended to set ``use_tmp_file`` to `False` when
+    using a ``staging_zone`` directory.
     """
     scns = job['resampled_scenes']
     objs = []
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -22,6 +22,7 @@
 """Trollflow2 plugins."""
 
 import os
+import pathlib
 from contextlib import contextmanager
 from logging import getLogger
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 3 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -189,9 +190,15 @@ def prepared_filename(fmat, renames):
...
     # tmp filenaming
     use_tmp_file = fmat.get('use_tmp_file', False)
+    use_tmp_dir = fmat.get("use_tmp_dir", False)
 
     if use_tmp_file:
-        filename = _get_temp_filename(directory, renames.keys())
+        if use_tmp_dir:
+            tmp_dir = pathlib.Path(use_tmp_dir)
+            of = pathlib.Path(orig_filename)
+            filename = os.fspath(tmp_dir / of.name)
```
mraspaud, line 199: Do I understand this correctly that the temporary file will not have a temporary name?

mraspaud: In which case I would recommend to have this option separated from the the `use_tmp_file` option

pnuu: Agree with @mraspaud. I think the behaviour should be something like this:
```python
if use_tmp_file:
    filename = _get_temp_filename(...)
elif use_tmp_dir:
    filename = os.path.join(use_tmp_dir, os.path.basename(orig_filename))
else:
    filename = orig_filename
```
And as the `use_tmp_dir` is used as the path, it should be renamed as `tmp_dir` or something.

pnuu (changes requested):

trollflow2/plugins/__init__.py
gerritholl: It is correct that the temporary file will not have a temporary name.  We need to retain the filename, or find a different solution to the issue described in #125.

I am not convinced that I agree with the separation.  A temporary file in a temporary directory is still a temporary file, it's just implemented differently.  Therefore, I don't think it makes sense to have `use_tmp_dir` without `use_tmp_file`.

mraspaud: ok, I can see a point in that, however the configuration needs to be clarified imo. So could we have something like this in the config:
```yaml
    use_tmp_file:
      - filepattern: {platform}_{product}  # should default to a random name like we had before
      - directory: /my/tmp/dir/ # should default to tempfile.gettempdir(), accepts "."
```
?

gerritholl: Thinking about this again, maybe it shouldn't be called `tmp_dir` at all, but something like `staging_zone`.  `tmp_dir` suggests a temporary directory is being created, but this is not the case, nor is it necessary.

gerritholl: @mraspaud @pnuu I have changed the logic.  The option is now called `staging_zone` and operates independently of `use_tmp_file` (although I'm not sure of a use case for using both, I have covered this case in the new unit test too).

trollflow2/plugins/__init__.py
```diff
@@ -257,11 +258,13 @@ def save_datasets(job):
     This is useful when other processes are waiting for the file to be present
     to start their work, but would crash on incomplete files.
 
-    If, in addition, the ``use_tmp_dir`` option is provided in the product
-    list, then the temporary file will be created in this directory without
+    If the ``staging_zone`` option is provided in the product
+    list, then the file will be created in this directory first, without
```
mraspaud, line 262: without changing the filename, you mean keeping either original or tmp file depending on `use_tmp_file`? (seems to be so if I understand the unit tests). Maybe this should be clarified?

trollflow2/plugins/__init__.py
```diff
@@ -257,11 +258,13 @@ def save_datasets(job):
...
     changing the filename.  This is useful for writers which write the filename
     to the headers, such as the SatPy ninjotiff and ninjogeotiff writers.  The
-    value of ``use_tmp_dir`` has no effect if ``use_tmp_file`` is False.
+    ``staging_zone`` directory must be on the same filesystem as
+    ``output_dir``.  It is recommended to set ``use_tmp_file`` to `False` when
+    using a ``staging_zone`` directory.
```
mraspaud, line 267: why?

trollflow2/plugins/__init__.py
gerritholl: Your interpretation is correct.  I clarified the documentation (see update).

trollflow2/plugins/__init__.py
gerritholl: Because otherwise there's little point in using `staging_zone`.  The whole point is such that the filename written to the headers by the ninjo(geo)tiff writers is meaningful, which it isn't if `use_tmp_file` is False.  Edited documentation for clarifying this.

mraspaud (approved):
LGTM

PR #135, opened 2022-01-21
Use UnsafeLoader when reading product list for process()

pnuu:
In #134 the config reading was refactored, and the `Loader` kwarg was accidentally changed when the product list is read.

 - [x] Tests added  
 - [x] Tests passed 

### DIFF
--- a/trollflow2/tests/test_launcher.py
+++ b/trollflow2/tests/test_launcher.py
@@ -601,8 +601,6 @@ def test_workers_initialized():
                     process("msg", fname, queue)
                 except StopIteration:
                     pass
-                else:
-                    raise
         finally:
             os.remove(fname)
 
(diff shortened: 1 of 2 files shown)


trollflow2/tests/test_launcher.py
```diff
@@ -581,6 +581,32 @@ def wait(job):
...
+        try:
+            with mock.patch("trollflow2.launcher.get_dask_client") as gdc:
+                # `get_dask_client()` is called just after config reading, so if we get there loading worked
+                gdc.side_effect = StopIteration
+                try:
+                    process("msg", fname, queue)
+                except StopIteration:
+                    pass
+                else:
+                    raise
```
mraspaud, line 605: codecov says this is not covered. Is this actually necessary in case the bug comes back?

pnuu: Apparently not. Removed.

mraspaud (approved):
LGTM

PR #51, opened 2019-09-24
Allow multiple bands tuple to be passed as a single composite

mraspaud:
The following syntax is now accepted in the YAML config file:

```yaml
    areas:
      baws:
        areaname: baws
        products:
          overview_sun:
          # ...
          ("1", "2"):
            productname: visible_channels
            formats:
              - format: nc
                writer: cf
                fname_pattern: "{start_time:%Y%m%d_%H%M}_{platform_name:s}_{productname:s}_{variant:s}.{format}"
```

PR #138, opened 2022-02-24
Make it possible to do eager saving

pnuu:
When saving multiple datasets to a single CF/NetCDF4 file using the syntax introduced in https://github.com/pytroll/trollflow2/pull/51 , I got random crashes resulting in `RuntimeError: NetCDF: Not a valid ID` within the XArray library. Some internet searching suggested that this is due to trying to use dimensions that have not yet been defined.

My solution to this is adding a config option that forces an eager saving instead of delaying the saving and calling `compute_writer_results()` afterwards. With this PR, I haven't seen this happen a single time over 50 consecutive runs.

 - [x] Tests added  
 - [ ] Tests passed 
 - [x] Fully documented

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -151,6 +151,13 @@ are several options:
 The ``fname_pattern`` can be a global setting at the top-level of the
 configuration, and overridden on area and product levels.
 
+It is possible to force the saving to be eager by defining
+``eager_writing: True`` in the product list. Eager saving means that
+each dataset are saved separately. The usage of this will most likely
+increase the processing time by a significant amount, but it is
+necessary until a `bug in XArray NetCDF4
+<https://github.com/pydata/xarray/issues/6300>`_ handling is fixed.
+
 Messaging for saved datasets
 ****************************
 
--- a/examples/pl.yaml
+++ b/examples/pl.yaml
@@ -57,6 +57,9 @@ product_list: &product_list
   # min_coverage: 20  # at least 20% coverage
   # use_tmp_file: False  # create temporary filename first
   # staging_zone: "/data/pytroll/tmp/staging_zone"  # create files here first
+  # Force eager writing and computation of datasets.
+  #   Required when saving several datasets to a single CF/NetCDF4 file.
+  # eager_writing: True
 
   areas:
     omerc_bb:
--- a/examples/pl.yaml
+++ b/examples/pl.yaml
@@ -58,7 +58,10 @@ product_list: &product_list
   # use_tmp_file: False  # create temporary filename first
   # staging_zone: "/data/pytroll/tmp/staging_zone"  # create files here first
   # Force eager writing and computation of datasets.
-  #   Required when saving several datasets to a single CF/NetCDF4 file.
+  #   Required when saving several datasets to a single CF/NetCDF4 file until
+  #   the bug in XArray NetCDF4 handling has been fixed. This option will most
+  #   likely increase the required processing time.
+  #   Issue on the XArray bug: https://github.com/pydata/xarray/issues/6300
   # eager_writing: True
 
   areas:
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -206,8 +206,11 @@ def prepared_filename(fmat, renames):
         yield orig_filename
 
 
-def save_dataset(scns, fmat, fmat_config, renames):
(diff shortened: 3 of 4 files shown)


pnuu:
Here's a short scipt using Satpy directly that demonstrates the random crashes:

```python
#!/usr/bin/env python

import glob

import numpy as np
from satpy import Scene
from satpy.writers import compute_writer_results

COMPUTE=False


def main():
    fnames = glob.glob(
        "/home/lahtinep/data/satellite/polar/ears_pps/*48965.nc")
    dsets = ['cma', 'ct', 'ctth_alti', 'ctth_pres', 'ctth_tempe']
    glbl = Scene(reader='nwcsaf-pps_nc', filenames=fnames)
    glbl.load(dsets)
    dtype_int16 = {'dtype': np.int16}
    encoding = {'cma': dtype_int16, 'ct': dtype_int16}
    res = glbl.save_datasets(writer='cf', filename="/tmp/pps_test.nc",
                             encoding=encoding, include_lonlats=True, compute=COMPUTE)
    compute_writer_results([res])


if __name__ == "__main__":
    main()
```

I'm trying to find an example that wouldn't require actual data.

pnuu:
And the corresponding version as `trollflow2.yaml`:
```yaml
product_list:

  output_dir:
    /tmp/
  fname_pattern:
    "{start_time:%Y%m%d_%H%M}_{platform_name}_{areaname}_EARS_PPS.nc"
  reader: nwcsaf-pps_nc
  subscribe_topics:
    - /test/ears/avhrr/pps/gatherer
  eager_writing: True

  areas:
    null:
      priority: 1
      areaname: swath
      products:
        ("cma", "ct", "ctth_alti", "ctth_pres", "ctth_tempe"):
          formats:
          - format: nc
            writer: cf
            encoding:
              cma:
                dtype: !!python/name:numpy.int16
              ct:
                dtype: !!python/name:numpy.int16
            include_lonlats: True


workers:
  - fun: !!python/name:trollflow2.plugins.create_scene
  - fun: !!python/name:trollflow2.plugins.load_composites
  - fun: !!python/name:trollflow2.plugins.resample
  - fun: !!python/name:trollflow2.plugins.save_datasets
```

pnuu:
Another pure-Satpy example using SEVIRI HRIT data:
```python

#!/usr/bin/env python

import glob

from satpy import Scene
from satpy.writers import compute_writer_results

COMPUTE = False


def main():
    fnames = glob.glob(
        "/home/lahtinep/data/satellite/geo/0deg/*202202031045*")
    dsets = ['VIS006', 'VIS008']
    glbl = Scene(reader='seviri_l1b_hrit', filenames=fnames)
    glbl.load(dsets)
    res = glbl.save_datasets(writer='cf', filename="/tmp/seviri_test.nc",
                             include_lonlats=True, compute=COMPUTE)
    if not COMPUTE:
        compute_writer_results([res])


if __name__ == "__main__":
    main()
```

pnuu:
And a version that fails without any Pytroll code:
```python
#!/usr/bin/env python

import numpy as np
import dask.array as da
import xarray as xr

COMPUTE = False
FNAME = "/tmp/xr_test.nc"


def main():
    y = np.arange(1000, dtype=np.uint16)
    x = np.arange(2000, dtype=np.uint16)

    # Write root
    root = xr.Dataset({}, attrs={'global': 'attribute'})
    written = [root.to_netcdf(FNAME, mode='w')]

    # Write first dataset
    data1 = xr.DataArray(da.random.random((y.size, x.size)), dims=['y', 'x'],
                         coords={'y': y, 'x': x})
    dset2 = xr.Dataset({'data1': data1})
    written.append(dset2.to_netcdf(FNAME, mode='a', compute=COMPUTE))

    # Write the same dataset with another name
    dset2 = xr.Dataset({'data2': data1})
    written.append(dset2.to_netcdf(FNAME, mode='a', compute=COMPUTE))

    if not COMPUTE:
        da.compute(written)


if __name__ == "__main__":
    main()
```

pnuu:
Created an issue to XArray: https://github.com/pydata/xarray/issues/6300

mraspaud (commented):
Thanks for adding this feature. I think it just needs adding in the general documentation that this option will most likely slow down processing significantly.

pnuu:
I added a note to the example configuration file. This will most likely be a temporary solution until https://github.com/pydata/xarray/issues/6300 is fixed.

PR #139, opened 2022-03-01
Fix warning issued when multiple sensors are provided

mraspaud:
This PR removes a log warning when the sensor is provided as a sequence of only one element.

 - [x] Tests added  
 - [x] Tests passed 


pnuu (approved):
LGTM

PR #140, opened 2022-03-09
Fix sunlight coverage check to return 100% when the data is fully lit

mraspaud:
The PR fixes so that the following data would report 100% sunlight coverage when check_pass is true:

Suomi-NPP_20220307_1323_euron1_natural_color

 - [x] Tests added  
 - [x] Tests passed


pnuu (approved):
LGTM. I guess the old `use_pass` was never even documented, the RtD page already speaks of `check_pass` :man_facepalming:

PR #146, opened 2022-04-26
Pop unknown keyword arguments in save_dataset

### LINKED ISSUE #145: save_dataset plugin mixes arguments for itself and for the writer
**Describe the bug**

The `trollflow2` plugin `save_dataset` can read arguments either directly from `product_list` or, if arguments vary between writers, from the `formats`-dictionary.  If we want to use different values of `staging_zone`, `output_dir`, `fname_pattern` for different writers, then we must configure them in the `formats`-dictionary.  In this case, `trollflow2` passes on those arguments to the satpy writer.  Although some writers (ninjotiff, geotiff) swallow any unknown arguments, others (netcdf, ninjogeotiff) are stricter and will fail on unknown arguments.

**To Reproduce**

```python
from satpy import Scene
from glob import glob
from queue import Queue
sc = Scene(
        filenames=seviri_files,
        reader=["seviri_l1b_hrit"])
sc.load(["IR_108"])
ls = sc.resample("eurol")

product_list = {
    "fname_pattern": "name.tif",
    "use_tmp_file": True,
    "staging_zone": "/tmp/a",
    "areas": {
        "eurol": {
            "products": {
                "IR_108": {
                    "productname": "IR108",
                    "formats": [
                        {"writer": "ninjogeotiff",
                         "ChannelID": 0,
                         "DataType": 0,
                         "PhysicUnit": "no",
                         "PhysicValue": "yes",
                         "SatelliteNameID": 0,
                         "output_dir": "/tmp/b",
                         }
                    ]
                }
            }
        }
    }
}

job = {}
job['resampled_scenes'] = {"eurol": ls}
job["input_mda"] = {}
job["produced_files"] = Queue()
save_datasets(job)
```

**Expected behavior**

Following "explicit is better than implimint", it would be better to have an interface where the configuration file clearly marks with arguments are for the satpy writer (`ChannelID`, `DataType`, `PhysicUnit`, `PhysicValue`, and `SatelliteNameID`), and which ones are for trollflow2 consumption only (`output_dir`).  Failing that, any arguments not expected by the writer should be removed before the writer is called.

**Actual results**

The code snippet fails, because trollflow2 passes `output_dir` to the writer, but the writer doesn't know what to do with this:

```
  File "/data/gholl/checkouts/protocode/mwe/trollflow2-save-dataset-args.py", line 42, in <module>
    save_datasets(job)
  File "/data/gholl/checkouts/trollflow2/trollflow2/plugins/__init__.py", line 284, in save_datasets
    obj = save_dataset(scns, fmat, fmat_config, renames, compute=eager_writing)
  File "/data/gholl/checkouts/trollflow2/trollflow2/plugins/__init__.py", line 232, in save_dataset
    obj = scns[fmat['area']].save_dataset(dsid,
  File "/data/gholl/checkouts/satpy/satpy/scene.py", line 1099, in save_dataset
    return writer.save_dataset(self[dataset_id],
  File "/data/gholl/checkouts/satpy/satpy/writers/__init__.py", line 814, in save_dataset
    return self.save_image(img, filename=filename, compute=compute, fill_value=fill_value, **kwargs)
  File "/data/gholl/checkouts/satpy/satpy/writers/ninjogeotiff.py", line 171, in save_image
    ntg = NinJoTagGenerator(
  File "/data/gholl/checkouts/satpy/satpy/writers/ninjogeotiff.py", line 318, in __init__
    raise ValueError("The following tags were not recognised: " +
ValueError: The following tags were not recognised: output_dir
```

**Environment Info:**

 - Trollflow2 Version: v0.13.4-8-gde82238
 - Satpy Version: v0.36.0-8-g20434bf3
 - Pyresample Version: v1.23.0-3-g1b14698

**Additional context**

This could be fixed by adding more lines to:

https://github.com/pytroll/trollflow2/blob/dbd4e5db1ebab5177e614d67fbc17b3c761a0a01/trollflow2/plugins/__init__.py#L219-L220

which works only as long as we can assume that no satpy writer uses any of those parameters.

A prettier solution would be to explicitly separate arguments for the plugin from arguments for the writer.

gerritholl:
When saving datasets, pop out all keyword arguments that are used by
the trollflow2 save_datasets plugin, but not by writers.

Add unit test to confirm this popping works.

 - [x] Closes #145 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -279,7 +279,7 @@ def save_datasets(job):
     Other arguments defined in the job list (either directly under
     ``product_list``, or under ``formats``) are passed on to the satpy writer.  The
     arguments ``use_tmp_file``, ``staging_zone``, ``output_dir``,
-    ``fname_pattern``, and ``dispatch`` are never passed to the reader.
+    ``fname_pattern``, and ``dispatch`` are never passed to the writer.
     """
     scns = job['resampled_scenes']
     objs = []
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -216,7 +216,7 @@ def save_dataset(scns, fmat, fmat_config, renames, compute=False):
         with prepared_filename(fmat, renames) as filename:
             res = fmat.get('resolution', DEFAULT)
             kwargs = fmat_config.copy()
-            # those keyword arguments are used by the trollflow2 plugin but not
+            # these keyword arguments are used by the trollflow2 plugin but not
             # by satpy writers
             for name in {"fname_pattern", "dispatch", "output_dir",
                          "use_tmp_file", "staging_zone"}:
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -216,8 +216,11 @@ def save_dataset(scns, fmat, fmat_config, renames, compute=False):
         with prepared_filename(fmat, renames) as filename:
             res = fmat.get('resolution', DEFAULT)
             kwargs = fmat_config.copy()
-            kwargs.pop('fname_pattern', None)
-            kwargs.pop('dispatch', None)
+            # those keyword arguments are used by the trollflow2 plugin but not
+            # by satpy writers
+            for name in {"fname_pattern", "dispatch", "output_dir",
+                         "use_tmp_file", "staging_zone"}:
+                kwargs.pop(name, None)
             if isinstance(fmat['product'], (tuple, list, set)):
                 kwargs.pop('format')
                 dsids = []
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -275,6 +275,11 @@ def save_datasets(job):
     it is recommended to set ``use_tmp_file`` to `False` when using a
     ``staging_zone`` directory, such that the filename written to the
     headers remains meaningful.
+
+    Other arguments defined in the job list (either directly under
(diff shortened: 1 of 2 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -272,6 +275,11 @@ def save_datasets(job):
     it is recommended to set ``use_tmp_file`` to `False` when using a
     ``staging_zone`` directory, such that the filename written to the
     headers remains meaningful.
+
+    Other arguments defined in the job list (either directly under
+    ``product_list``, or under ``formats``) are passed on to the satpy writer.  The
+    arguments ``use_tmp_file``, ``staging_zone``, ``output_dir``,
+    ``fname_pattern``, and ``dispatch`` are never passed to the reader.
```
pnuu, line 282: Are never passed to the reader or writer?

gerritholl: That should be writer.  Fixed in 9df0420, thanks.

trollflow2/plugins/__init__.py
```diff
@@ -216,8 +216,11 @@ def save_dataset(scns, fmat, fmat_config, renames, compute=False):
         with prepared_filename(fmat, renames) as filename:
             res = fmat.get('resolution', DEFAULT)
             kwargs = fmat_config.copy()
-            kwargs.pop('fname_pattern', None)
-            kwargs.pop('dispatch', None)
+            # those keyword arguments are used by the trollflow2 plugin but not
```
pnuu, line 219:
```suggestion
            # these keyword arguments are used by the trollflow2 plugin but not
```

PR #147, opened 2022-04-28
Update native resampler handling

pnuu:
Trollflow2 was still using the deprecated `scn.min_area()` and `scn.max_area()` methods for `null` areas. This PR switches over to the modern `scn.coarsest_area()` and `scn.finest_area()` and does some refactoring. Also the old config entries `use_min_area` and `use_max_area` are deprecated in favor of `use_coarsest_area` and `use_finest_area`.

 - [x] Tests updated  
 - [x] Tests passed 

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -128,30 +128,32 @@ def resample(job):
                                      conf)
         LOG.debug('Resampling to %s', str(area))
         if area == 'None':
-            minarea = get_config_value(product_list,
-                                       '/product_list/areas/' + str(area),
-                                       'use_min_area')
-            maxarea = get_config_value(product_list,
-                                       '/product_list/areas/' + str(area),
-                                       'use_max_area')
-            native = conf.get('resampler') == 'native'
-            if minarea is True:
-                job['resampled_scenes'][area] = scn.resample(scn.min_area(),
-                                                             **area_conf)
-            elif maxarea is True:
-                job['resampled_scenes'][area] = scn.resample(scn.max_area(),
-                                                             **area_conf)
-            elif native:
-                job['resampled_scenes'][area] = scn.resample(resampler='native')
-            else:
-                # The composites need to be created for the saving to work
-                if not set(scn.keys()).issuperset(scn.wishlist):
-                    LOG.debug("Generating composites for 'null' area (satellite projection).")
-                    scn.load(scn.wishlist, generate=True)
-                job['resampled_scenes'][area] = scn
+            job['resampled_scenes'][area] = _resample_null_area(product_list, scn, conf, area_conf)
         else:
             LOG.debug("area: %s, area_conf: %s", area, str(area_conf))
             job['resampled_scenes'][area] = scn.resample(area, **area_conf)
+
+
+def _resample_null_area(product_list, scn, conf, area_conf):
+    minarea = get_config_value(product_list,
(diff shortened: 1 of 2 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -128,32 +128,46 @@ def resample(job):
...
-                                                             **area_conf)
-            elif native:
-                job['resampled_scenes'][area] = scn.resample(resampler='native')
-            else:
-                # The composites need to be created for the saving to work
-                if not set(scn.keys()).issuperset(scn.wishlist):
-                    LOG.debug("Generating composites for 'null' area (satellite projection).")
-                    scn.load(scn.wishlist, generate=True)
-                job['resampled_scenes'][area] = scn
+            job['resampled_scenes'][area] = _resample_null_area(product_list, scn, conf, area_conf)
```
gerritholl, line 131: Is there any documentation for the `area == 'None'` behaviour?

gerritholl (approved):
Looks good to me, just one remark regarding documentation.

trollflow2/plugins/__init__.py
pnuu: The only place I could find is in https://github.com/pytroll/trollflow2/blob/main/examples/pl.yaml#L114

We'd really need to add some proper documentation and not hide things only in (commented-out in this case) example configuration files.

gerritholl: In the example it says `null` though, not `None`?

pnuu: In yaml `null` is used to denote `None` of Python. I don't remember why in the code we change `None` to `'None'` though...

mraspaud: do we have a test for that?

trollflow2/plugins/__init__.py
```diff
@@ -128,32 +128,46 @@ def resample(job):
...
+        return scn.resample(resampler='native')
+
+    # The composites need to be created for the saving to work
+    if not set(scn.keys()).issuperset(scn.wishlist):
+        LOG.debug("Generating composites for 'null' area (satellite projection).")
+        scn.load(scn.wishlist, generate=True)
+    return scn
+
+
+def _get_native_area(product_list, area_name, legacy='use_min_area'):
```
mraspaud, line 155: I'm not convince `area_name` is the right parameter name here... would you mind changing it to, say, `area_grouping_method`?
Same for `legacy`, maybe `legacy_grouping_method`?

trollflow2/tests/test_trollflow2.py
```diff
@@ -830,25 +830,25 @@ def test_minmax_area(self):
...
                                   resampler="nearest",
                                   reduce_data=True,
                                   cache_dir=None,
                                   mask_area=False,
                                   epsilon=0.0) in
                         scn.resample.mock_calls)
-        del product_list['product_list']['areas']['None']['use_min_area']
-        product_list['product_list']['areas']['None']['use_max_area'] = True
+        del product_list['product_list']['areas']['None']['use_coarsest_area']
+        product_list['product_list']['areas']['None']['use_finest_area'] = True
```
mraspaud, line 848: I think the `legacy` options should be tested too, making sure the warning is issued.

mraspaud (changes requested):
Thanks for updating! much needed.

pnuu:
Superseded by #157. Closing.

PR #150, opened 2022-05-20
Adapt remote (fsspec) file reading for newer pytroll-collector changes

mraspaud:
Recent changes in the pytroll-collector package regarding formatting of message for remote filesystems require trollflow2 to adapt. This PR implements the adaptation.

 - [x] Tests added

### DIFF
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -39,7 +39,6 @@ from contextlib import suppress
 from datetime import datetime
 from logging import getLogger
 from queue import Empty
-from urllib.parse import urlparse
 
 import yaml
 from trollflow2.dict_tools import gen_dict_extract, plist_iter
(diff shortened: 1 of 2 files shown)


trollflow2/launcher.py
```diff
@@ -266,7 +266,7 @@ def _extract_filenames(msg):
 
     If the message contains a `filesystem` item, use fsspec to decode it.
     """
-    filenames = [urlparse(uri).path for uri in gen_dict_extract(msg.data, 'uri')]
+    filenames = [uri.split("::")[0] for uri in gen_dict_extract(msg.data, 'uri')]
```
pnuu, line 269: Wouldn't this return just `'simplecache'` if the URI is something like `simplecache::s3://noaa-goes17/ABI-L1b-RadF/2021/195/20/*_G17_s20211952030*`?

pnuu (commented):
Just one question.

trollflow2/launcher.py
adybbroe: Seems reasonable @pnuu , perhaps we should add a test for that then?

mraspaud: Looking at it now

pnuu (approved):
LGTM

PR #151, opened 2022-06-03
Add checking for incoming data age

pnuu:
With this PR it is possible to discard "too old" data before it is processed.

My use-case is the occasional data outages in Eumetacast, where the data are often sent late to make the data series intact. This causes the processing queue to stay full for hours (depending on the outage duration), and delays the processing of current data. With this feature it would be possible to discard the old data.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -283,6 +283,11 @@ discarded.  If the configured item isn't in the input message
 metadata, a warning is printed and the processing continues.  The best
 place for this plugin is before any data handling is done.
 
+In addition to equality checks, a negative integer ``start_time`` can be
+defined to denote the maximum age of the accepted data. For example
+``start_time: -60`` means that all data older than 60 minutes will be
+discarded.
+
 Options:
  - ``check_metadata: null`` - A dictionary of metadata names and list(s)
    of values that need to match.  By default (``null``) all scenes are
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -285,7 +285,7 @@ place for this plugin is before any data handling is done.
 
 In addition to equality checks, ``start_time`` can be used to check
 for data that are either too old or too new to be processed. To skip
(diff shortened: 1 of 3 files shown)


mraspaud:
Maybe a stupid question, but should this not be done earlier in the chain? like trollmoves or segment gatherer?

pnuu:
I was thinking about that. And if I made it so that there'd be option to process _only_ old ones, we could use the same collectors and have a "history processor" as a separate Trollflow2 entity, but still have only one collector. I think this is definitely the correct place for the filtering.

pnuu:
This now does what I was thinking about, the logic might not be that clear. Even I got my brains in a knot defining the tests...

mraspaud (changes requested):
Looks good, just a couple of suggestions.

mraspaud:
This is an interesting problem. At the moment we don't have a priority system for the data, but if we had we could lower it for old data. That would allow for old data to be processed during the idle times...
Having a priority system is something I thought of when seeing how we could use trollflow2 in combination with dask distributed. One instance of trollflow2 would handle all different data types... but yeah, nothing we need to implement now :)

doc/source/index.rst
```diff
@@ -287,6 +287,12 @@ discarded.  If the configured item isn't in the input message
 metadata, a warning is printed and the processing continues.  The best
 place for this plugin is before any data handling is done.
 
+In addition to equality checks, ``start_time`` can be used to check
+for data that are either too old or too new to be processed. To skip
+accepted data older than a maximum age (in minutes) , use a negative integer:
+``start_time: -60``. Similarly positive integer skips the processing of
+data newer than than the defined time.
```
pnuu, line 294:
```suggestion
data newer than the defined time.
```

mraspaud (approved):
LGTM

PR #152, opened 2022-06-08
Do not crash when computing coverage of a dynamic area def

mraspaud:
Computing the coverage of a dynamic area def does not work at the moment since the bbox can not be computed (understandably).
This fix gives a coverage of 100% for these types of areas.

 - [x] Tests added  
 - [x] Tests passed 


pnuu (approved):
LGTM

PR #154, opened 2022-07-15
Pass arbitrary params to Scene.load

### LINKED ISSUE #153: Pass parameters to Scene.load
## Feature Request

In Satpy, some readers can interpret parameters passed to `Scene.load()`.  For example, [readers for some geostationary imagers](https://satpy.readthedocs.io/en/stable/readers.html#load-data) can understand `pad_data` and `upper_right_corner`.  The trollflow2 `load_composites` plugin does not pass any parameters except `resolution` and `generate`, nor does it allow any configuration for the user to pass others:

https://github.com/pytroll/trollflow2/blob/265df393cd315862289e27ed98f0c4d4df49a8c8/trollflow2/plugins/__init__.py#L102

A user may want to write images in native geostationary projection using trollflow2, but still flipping the image so that the resulting images follow the convention of having the north "up".  This is not currently possible with trollflow2.

I would like that the user can define in the trollflow2 YAML configuration the parameters to be passed to `Scene.load()` by the `load_composites` plugin.  I'm not sure where this would be defined.  Maybe a `reader_load_args` field directly under the `product_list`, analogous to the existing `reader_kwargs`, or maybe within the `reader` field in the `product_list`.

Depends on the chosen solution, but it should be possible to choose a solution that does not affect existing user workflow.

**Additional context**

One could resample to one of the nominal satellite-projection areas defined in `satpy/etc/areas.yaml`, but this is computationally much more expensive than just flipping.

gerritholl:
Allow to configure arbitrary extra parameters to be passed to
Scene.load.

 - [x] Closes #153  
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -742,7 +742,6 @@ class TestLoadComposites(TestCase):
 
     def test_load_composites_with_custom_args(self):
         """Test loading with arbitrary additional arguments."""
-        """Test loading composites with different resolutions."""
         from trollflow2.plugins import load_composites, DEFAULT
         scn = _get_mocked_scene_with_properties()
         self.product_list['product_list']['scene_load_kwargs'] = {"upper_right_corner": "NE"}
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -740,6 +740,18 @@ class TestLoadComposites(TestCase):
         scn.load.assert_any_call({'cloudtype', 'ct', 'cloud_top_height'}, resolution=1000, generate=True)
         scn.load.assert_any_call({'cloud_top_height'}, resolution=500, generate=True)
 
+    def test_load_composites_with_custom_args(self):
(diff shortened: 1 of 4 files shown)


trollflow2/tests/test_trollflow2.py
```diff
@@ -740,6 +740,18 @@ def test_load_composites_with_different_resolutions(self):
         scn.load.assert_any_call({'cloudtype', 'ct', 'cloud_top_height'}, resolution=1000, generate=True)
         scn.load.assert_any_call({'cloud_top_height'}, resolution=500, generate=True)
 
+    def test_load_composites_with_custom_args(self):
+        """Test loading with arbitrary additional arguments."""
+        """Test loading composites with different resolutions."""
```
mraspaud, line 745: Which one is it? ;)

mraspaud (changes requested):
Looks good! just one suggestion, then it's good to merge

mraspaud (approved):
LGTM!

PR #158, opened 2022-09-07
Add a plugin to upload the generated images to S3

pnuu:
This PR adds a plugin that updloads the generated files to S3. Optionally also deletes the local files.

 - [x] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -892,3 +892,32 @@ def _product_meets_min_valid_data_fraction(
     LOG.debug(f"Found {rel_valid:%}>{min_frac:%}, keeping "
               f"{prod_name:s} for area {area_name:s} in the worklist")
     return True
+
+
+def s3_uploader(job):
+    """Upload data to S3 and update the filenames.
+
+    Optionally also delete the files after the upload.
+    """
+    s3_config = job['product_list']['product_list'].get('s3_config', {}).copy()
+    delete_files = s3_config.pop('delete_files', False)
+    s3 = _get_s3_connection(s3_config)
+
+    for fmt, fmt_config in plist_iter(job['product_list']['product_list']):
+        local_fname = fmt_config['filename']
+        s3_uri = fmt['filename'].replace(fmt['output_dir'], fmt['s3_config']['target'])
+        _s3_upload(s3, local_fname, s3_uri)
+        fmt_config['filename'] = s3_uri
+        if delete_files:
+            os.remove(local_fname)
+
+
+def _get_s3_connection(s3_config):
+    from s3fs import S3FileSystem
+
+    _ = s3_config.pop('target')
+    return S3FileSystem(**s3_config)
+
+
+def _s3_upload(s3, local_fname, s3_uri):
+    s3.put(local_fname, s3_uri)
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -1679,7 +1679,7 @@ class TestFilePublisher(TestCase):
             pub = FilePublisher()
             product_list = self.product_list.copy()
             product_list['product_list']['publish_topic'] = '/{areaname}/{productname}'
-            topics = self._create_filenames_and_topics(job)
+            topics = _create_filenames_and_topics(job)
 
             pub(job)
             message.assert_called()
@@ -1761,30 +1761,9 @@ class TestFilePublisher(TestCase):
         pub = FilePublisher()
         product_list = self.product_list.copy()
         product_list['product_list']['publish_topic'] = '/static_topic'
-        topics = self._create_filenames_and_topics(job)
+        topics = _create_filenames_and_topics(job)
         pub(job)
         return pub, topics
-
-    @staticmethod
-    def _create_filenames_and_topics(job):
-        """Create the filenames and topics for *job*."""
-        from trollflow2.dict_tools import plist_iter
-        from trollsift import compose
-        import os.path
-
-        topic_pattern = job['product_list']['product_list']['publish_topic']
-        topics = []
-
-        for fmat, fmat_config in plist_iter(job['product_list']['product_list'],
-                                            job['input_mda'].copy()):
-            fname_pattern = fmat['fname_pattern']
-            filename = compose(os.path.join(fmat['output_dir'],
-                                            fname_pattern), fmat)
-            fmat.pop('format', None)
-            fmat_config['filename'] = filename
-            topics.append(compose(topic_pattern, fmat))
-
-        return topics
 
     def test_filepublisher_kwargs(self):
         """Test filepublisher keyword argument usage."""
@@ -1907,6 +1886,27 @@ class TestFilePublisher(TestCase):
         nb_.stop.assert_not_called()
         pub.stop()
         nb_.stop.assert_called_once()
+
+
+def _create_filenames_and_topics(job):
+    """Create the filenames and topics for *job*."""
+    from trollflow2.dict_tools import plist_iter
+    from trollsift import compose
+    import os.path
+
+    topic_pattern = job['product_list']['product_list']['publish_topic']
+    topics = []
+
+    for fmat, fmat_config in plist_iter(job['product_list']['product_list'],
+                                        job['input_mda'].copy()):
+        fname_pattern = fmat['fname_pattern']
+        filename = compose(os.path.join(fmat['output_dir'],
+                                        fname_pattern), fmat)
+        fmat.pop('format', None)
+        fmat_config['filename'] = filename
+        topics.append(compose(topic_pattern, fmat))
+
+    return topics
 
 
 class FakeScene(dict):
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -2015,5 +2015,145 @@ def test_persisted(sc_3a_3b):
     assert not sc_3a_3b["another"].attrs.get("persisted")
 
 
+yaml_test_s3_uploader_plain = """
+product_list:
+  output_dir: /tmp
+  publish_topic: /topic
+  s3_config:
+    target: s3://bucket-name
+  fname_pattern:
+    "{start_time:%Y%m%d_%H%M}_{platform_name}_{areaname}_{productname}.{format}"
+
+  areas:
+    euro4:
+      areaname: euro4
+      products:
+        airmass:
+          productname: airmass
+          formats:
+          - format: tif
+            writer: geotiff
+        natural_with_ir:
+          productname: natural_with_colorized_ir_clouds
+          formats:
+          - format: tif
+            writer: geotiff
+
+"""
+
+s3_config_public = {
+    "target": "public-bucket-name",
+    "anon": True,
+}
+s3_config_more_settings = {
+    "client_kwargs": {
+        "endpoint_url": "https://server.url",
+    },
+    "key": "accesskey",
+    "secret": "secretkey",
+}
+
+
+def test_s3_uploader_update_filenames():
+    """Ensure that filenames are updated when transfer is made to S3."""
+    from yaml import UnsafeLoader
+    from trollflow2.plugins import s3_uploader
+    from trollflow2.dict_tools import plist_iter
+
+    product_list = read_config(raw_string=yaml_test_s3_uploader_plain, Loader=UnsafeLoader)
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('s3fs.S3FileSystem'):
+        s3_uploader(job)
+    for fmt, _ in plist_iter(job['product_list']['product_list']):
+        assert fmt['filename'].startswith('s3://bucket-name/')
+
+
+def test_s3_uploader_plain_connection():
+    """Test that S3 connection is initialized correctly for plain config file."""
+    from yaml import UnsafeLoader
+    product_list = read_config(raw_string=yaml_test_s3_uploader_plain, Loader=UnsafeLoader)
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('s3fs.S3FileSystem') as s3filesystem:
+        from trollflow2.plugins import s3_uploader
+
+        s3_uploader(job)
+        s3filesystem.assert_called_once_with()
+
+
+def test_s3_uploader_anon_connection():
+    """Test that S3 connection is initialized correctly for anonymous target."""
+    from yaml import UnsafeLoader
+    product_list = read_config(raw_string=yaml_test_s3_uploader_plain, Loader=UnsafeLoader)
+    product_list['product_list']['s3_config'] = s3_config_public.copy()
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('s3fs.S3FileSystem') as s3filesystem:
+        from trollflow2.plugins import s3_uploader
+
+        s3_uploader(job)
+        s3filesystem.assert_called_once_with(anon=True)
+
+
+def test_s3_uploader_complex_connection():
+    """Test that S3 connection is initialized correctly for more connection options."""
+    from yaml import UnsafeLoader
+    product_list = read_config(raw_string=yaml_test_s3_uploader_plain, Loader=UnsafeLoader)
+    product_list['product_list']['s3_config'].update(s3_config_more_settings)
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('s3fs.S3FileSystem') as s3filesystem:
+        from trollflow2.plugins import s3_uploader
+
+        s3_uploader(job)
+        s3filesystem.assert_called_once_with(**s3_config_more_settings)
+
+
+def test_s3_uploader_put():
+    """Test that S3 upload is called correctly."""
+    from yaml import UnsafeLoader
+    product_list = read_config(raw_string=yaml_test_s3_uploader_plain, Loader=UnsafeLoader)
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('s3fs.S3FileSystem') as s3filesystem:
+        from trollflow2.plugins import s3_uploader
+
+        s3_uploader(job)
+
+        assert mock.call(
+            "/tmp/20190217_0600_NOAA-15_euro4_airmass.tif",
+            "s3://bucket-name/20190217_0600_NOAA-15_euro4_airmass.tif"
+        ) in s3filesystem.return_value.put.mock_calls
+        assert mock.call(
+            "/tmp/20190217_0600_NOAA-15_euro4_natural_with_colorized_ir_clouds.tif",
+            "s3://bucket-name/20190217_0600_NOAA-15_euro4_natural_with_colorized_ir_clouds.tif"
+        ) in s3filesystem.return_value.put.mock_calls
+
+
+def test_s3_uploader_delete_files():
+    """Test that files are deleted after S3 transfer."""
+    from yaml import UnsafeLoader
+    product_list = read_config(raw_string=yaml_test_s3_uploader_plain, Loader=UnsafeLoader)
+    product_list['product_list']['s3_config']['delete_files'] = True
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('s3fs.S3FileSystem'):
+        with mock.patch('os.remove') as os_remove:
+            from trollflow2.plugins import s3_uploader
+
+            s3_uploader(job)
+
+            assert mock.call("/tmp/20190217_0600_NOAA-15_euro4_airmass.tif") in os_remove.mock_calls
+            assert mock.call(
+                "/tmp/20190217_0600_NOAA-15_euro4_natural_with_colorized_ir_clouds.tif") in os_remove.mock_calls
+
+
 if __name__ == '__main__':
     unittest.main()


mraspaud:
@pnuu thanks for the work in this PR.
I thought a little about this, and I think though that what is being added here is in fact a dispatching mechanism, for which we have a process in trollmoves: https://github.com/pytroll/trollmoves/blob/main/trollmoves/dispatcher.py
I think the benefit with the dispatcher is that no time is lost dispatching/uploading the file inside the trollflow run...

pnuu:
Ok, I've tested this PR with proper reading and saving, and it works.

There is just the `trollflow2.launcher.check_results` function that is logging `ERROR` messages due to the files are not available anymore after S3 upload and subsequent deletion. Any thoughts what to do about this @mraspaud?

pnuu:
Still some work needed in the messaging, the published messages have URIs like this: `/tmp/s3:/bucket-name/20220908_1050_Meteosat-10_EPSG3035_wv73.tif`. So the temporary save path is there, and also a slash is missing after the scheme.

pnuu:
With the latest commit the messages are now working. Still need to figure out what to do with `trollflow2.launcher.check_results`.

pnuu:
> @pnuu thanks for the work in this PR. I thought a little about this, and I think though that what is being added here is in fact a dispatching mechanism, for which we have a process in trollmoves: https://github.com/pytroll/trollmoves/blob/main/trollmoves/dispatcher.py I think the benefit with the dispatcher is that no time is lost dispatching/uploading the file inside the trollflow run...

Hadn't noticed this comment. The dispatcher would work when running on a (virtual) server with multiple tasks, but not when running in a container where there's supposed to be only a single process running and the saving is done inside the container. Using a shared volume might work, but that would also add a new container and set of configuration files.

pnuu:
This now uses `trollmoves.movers.S3Mover` instead of direct `s3fs` calls.

pnuu:
It seems that the upload is done to `s3://<bucket name>/<image name>/<image name>`...

pnuu:
Ok, it was a usage error that was revealed when uploading to real S3. The filename was interpreted as a prefix ("directory") by `botocore`. Fixed with https://github.com/pytroll/trollflow2/pull/158/commits/5d53219409ae5e5f0c5b13429ff83001de33941e

pnuu:
> Ok, I've tested this PR with proper reading and saving, and it works.
> 
> There is just the `trollflow2.launcher.check_results` function that is logging `ERROR` messages due to the files are not available anymore after S3 upload and subsequent deletion. Any thoughts what to do about this @mraspaud?

I'm now thinking that maybe the S3 uploading and local file deletion should be separate plugins :thinking: The other options seem more complicated.

mraspaud:
I thought a bit more about this, to try to keep things clean and separated. Could we have a custom class for taking care of files? Something that mimics a python file object for example, and that uploads the file on `close` for example?

pnuu:
I think this should work for what I need. I'll add a test for this tomorrow unless there's something more urgent happening.

trollflow2/launcher.py
```diff
@@ -84,7 +88,7 @@ def get_test_message(test_message_file):
     return msg
 
 
-def check_results(produced_files, start_time, exitcode):
+def check_results(produced_files, start_time, exitcode, remote_filesystem=None):
```
mraspaud, line 91: Instead of passing the filesystem here, could we have `produced_files` be a list of FSFiles (or subclass of it) so that it contains both filename and filesystem?
Also, having a subclass of FSFile would allow to add a check_result method to it, which would be nice.

trollflow2/plugins/__init__.py
```diff
@@ -40,6 +40,10 @@
 from pyresample.boundary import AreaDefBoundary, Boundary
 from pyresample.area_config import AreaNotFound
 from rasterio.enums import Resampling
+try:
+    from trollmoves.movers import S3Mover
+except ImportError:
+    S3Mover = None
```
mraspaud, line 46: Could we make the s3 uploader plugin a separate module so that we can just import this without the try except?

trollflow2/tests/test_launcher.py
```diff
@@ -773,6 +773,32 @@ def qsize(self):
...
+    assert "Missing file" in caplog.text
+    assert "files produced nominally" not in caplog.text
+
+    produced_files = FakeQueue(10, 13)
+    with mock.patch("trollflow2.launcher.S3FileSystem") as S3FileSystem:
+        check_results(produced_files, start_time, 0, remote_filesystem=remote_filesystem)
+    assert mock.call('s3://bucket-name/file10') in S3FileSystem.return_value.stat.mock_calls
+    assert mock.call('s3://bucket-name/file11') in S3FileSystem.return_value.stat.mock_calls
+    assert mock.call('s3://bucket-name/file12') in S3FileSystem.return_value.stat.mock_calls
+
```
mraspaud, line 801: These should be separate tests, the function here is already way too long :)

trollflow2/tests/test_trollflow2.py
```diff
@@ -2015,5 +2049,88 @@ def fake_persist(*args):
...
+          formats:
+          - format: tif
+            writer: geotiff
+        natural_with_ir:
+          productname: natural_with_colorized_ir_clouds
+          formats:
+          - format: tif
+            writer: geotiff
+
+"""
```
mraspaud, line 2076: maybe put the s3 tests in another test module, eg `test_s3_upload.py`?

mraspaud (changes requested):
Thanks for adding this feature. I think we need to restructure things a bit to avoid ugly import patterns and a better encapsulation to make the code easier to read.

doc/source/index.rst
```diff
@@ -368,6 +389,10 @@ Example
       - avhrr-3
 
     min_coverage: 25
+    s3_config:
+      target: s3://name-of-the-bucket/
+      delete_files: true
```
gerritholl, line 394: Do we still need `output_dir` when we use `s3_config` like this?  It would just be a temporary directory like `staging_zone`?  It would seem that this could be optional if `delete_files` is `True`, and that if the user has not defined what temporary directory to use, trollflow2 could create its own with `tempfile.TemporaryDirectory`?

trollflow2/tests/test_trollflow2.py
```diff
@@ -2015,5 +2049,88 @@ def fake_persist(*args):
...
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('trollflow2.plugins.S3Mover') as S3Mover:
+        from trollflow2.plugins import s3_uploader
+
+        s3_uploader(job)
+
+        assert mock.call(
+            "/tmp/20190217_0600_NOAA-15_euro4_airmass.tif",
```
gerritholl, line 2108: Should this use `tmp_path`?

trollflow2/tests/test_trollflow2.py
```diff
@@ -2015,5 +2049,88 @@ def fake_persist(*args):
...
+    product_list['product_list']['s3_config']['delete_files'] = True
+    job = {"product_list": product_list, "input_mda": input_mda.copy()}
+    _ = _create_filenames_and_topics(job)
+
+    with mock.patch('trollflow2.plugins.S3Mover') as S3Mover:
+        from trollflow2.plugins import s3_uploader
+
+        s3_uploader(job)
+
+        assert S3Mover.return_value.move.call_count == 2
```
gerritholl, line 2132: Where is this actually checking that files were deleted?

gerritholl (commented):
Some questions and suggestions, but nothing critical.

trollflow2/launcher.py
```diff
@@ -44,6 +44,10 @@
 from trollflow2.dict_tools import gen_dict_extract, plist_iter
 from trollflow2.logging import setup_queued_logging
 from trollflow2.plugins import AbortProcessing
+try:
+    from s3fs import S3FileSystem
+except ImportError:
+    S3FileSystem = None
```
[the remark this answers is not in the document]
pnuu: Done in https://github.com/pytroll/trollflow2/pull/158/commits/ec48ad3d9a1b6bfadbb566cb0bd77f5896ced82c

trollflow2/tests/test_launcher.py
pnuu: I'd split the test, but I can't figure out where `tmp_path` is defined (guess it's a pytest fixture?) and how to pass it to `FakeQueue` without the now-used closure. I guess one option would be to use `tempfile` module, but I take it there's a reason to use the hidden `tmp_path` so don't want to go that way either.

trollflow2/tests/test_trollflow2.py
pnuu: No, the user has defined `/tmp/` as the output directory. Nothing is written here, we just check that the mover is called with correct values.

trollflow2/launcher.py
```diff
@@ -120,6 +127,32 @@ def check_results(produced_files, start_time, exitcode):
...
+
+def _check_s3_file(saved_file, remote_filesystem):
+    if S3FileSystem is None:
+        raise ImportError("'s3fs' is required for S3 file check.")
+    s3 = S3FileSystem()
+    remote_file = os.path.join(remote_filesystem, os.path.basename(saved_file))
+    if s3.stat(remote_file)['size'] == 0:
+        LOG.error("Empty file detected: %s", remote_file)
+        return True
+    return False
```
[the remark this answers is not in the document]
pnuu: That's because the return value stored in `check_results()` is `error_detected`.

doc/source/index.rst
pnuu: The `output_dir` is used by the `save_datasets` plugin, and defaults to current directory. The `s3_uploader` just gets the filenames from the product list and uploads them to the target bucket. Making the `s3_config` to affect the `save_datasets` would make the usage quite complex. At least complex to explain.

trollflow2/tests/test_trollflow2.py
pnuu: The docstring is a bit out-of-sync. The deletion isn't tested directly, just that the `.move()` method of `S3Mover` is called. I'll update the docstring.

pnuu: Updated in https://github.com/pytroll/trollflow2/pull/158/commits/9d103ea3f78bb6341cee3b0dfaf3182b6388a758

trollflow2/plugins/__init__.py
pnuu: I moved the import inside the plugin in https://github.com/pytroll/trollflow2/pull/158/commits/f45e03e0d7175d349bf6703ea303db1c32c0b655 so the try/except isn't needed.

trollflow2/launcher.py
```diff
@@ -96,12 +100,15 @@ def check_results(produced_files, start_time, exitcode):
...
-                if os.path.getsize(saved_file) == 0:
-                    LOG.error("Empty file detected: %s", saved_file)
-                    error_detected = True
+                error_detected = _check_file(saved_file, remote_filesystem)
             except FileNotFoundError:
                 LOG.error("Missing file: %s", saved_file)
                 error_detected = True
+            except NotImplementedError as err:
+                LOG.error(err)
+                error_detected = True
```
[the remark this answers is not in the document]
pnuu: I think the error message in `trollflow2.launcher._check_file()` is clear enough. Or is it?

trollflow2/launcher.py
pnuu: This isn't actually the filesystem (like 's3' or any fs object), it's the URI to the S3 bucket. The filenames to `check_results()` are coming from `save_datasets()` plugin, which doesn't know about the `s3_uploader()`. And I think it shouldn't either.

I'll clarify the argument naming.

pnuu: Argument name changed in https://github.com/pytroll/trollflow2/pull/158/commits/27f62d05b100c45d1aec3171cb3fe498a98e4a17

trollflow2/launcher.py
```diff
@@ -120,6 +127,32 @@ def check_results(produced_files, start_time, exitcode):
                      f"{elapsed!s}", extra={"time": elapsed})
 
 
+def _check_file(saved_file, remote_filesystem):
+    if remote_filesystem is None:
+        return _check_local_file(saved_file)
+    if remote_filesystem.startswith('s3'):
+        return _check_s3_file(saved_file, remote_filesystem)
+    raise NotImplementedError("File check not impleneted for remote filesystem %s" % remote_filesystem)
```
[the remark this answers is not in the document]
pnuu: Until further notice, yes, nobody's using other than local or S3 filesystems :grin:

trollflow2/tests/test_trollflow2.py
pnuu: Split to `test_s3_plugins.py` in https://github.com/pytroll/trollflow2/pull/158/commits/3505b3a071986eeef59295e4fee47861f1d1cb7f

trollflow2/tests/test_launcher.py
pnuu: Split to separate tests in https://github.com/pytroll/trollflow2/pull/158/commits/ea049f1b94240f5199b9139a51b4ad6f1ea57ae0

The `FakeQueue` class didn't need a specific init interface, so the closure was never needed.

trollflow2/launcher.py
```diff
@@ -213,7 +247,14 @@ def _run_product_list_on_messages(self, messages, target_fun, process_class):
...
             except AttributeError:
                 exitcode = 0
-            check_results(produced_files_queue, start_time, exitcode)
+            check_results(produced_files_queue, start_time, exitcode,
+                          remote_filesystem=remote_filesystem)
+
+    def _get_remote_filesystem(self):
+        remote_filesystem = None
+        if 's3_config' in self.product_list['product_list']:
+            remote_filesystem = self.product_list['product_list']['s3_config']['target']
```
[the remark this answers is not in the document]
pnuu: Same as in https://github.com/pytroll/trollflow2/pull/158#discussion_r1011443142 . The filenames don't contain information where they were transferred after the `save_datasets()` plugin wrote them.

pnuu:
I think this is now closer to ready, and all comments should have been handled in a way or the other.

pnuu:
WTH did that `feature-s3-upload-plugin` came from?! It's now in 18 random places in `test_trollflow2.py` :man_facepalming:

doc/source/index.rst
mraspaud, line 394: I would like to avoid having a configuration for s3 here, because if we decide to add other protocols, it will require still other config sections.
My wish is to keep things simple in the config, and in the code.
So as an alternative, could we just use the `output_dir` parameter for specifying the target protocol? Like
`output_dir: s3://name-of-the-bucket/some/dir/`?
Then, we can use the `staging_zone` parameter to set where the files are saved locally before the transfer.

trollflow2/launcher.py
```diff
@@ -85,7 +85,7 @@ def get_test_message(test_message_file):
     return msg
 
 
-def check_results(produced_files, start_time, exitcode):
+def check_results(produced_files, start_time, exitcode, remote_filesystem_uri=None):
```
mraspaud, line 88: This starts to get a bit complicated here with an extra parameter that refers to the produced files. Also, later in this module, the `_check_file` function is created, that contains if-statements to check all the cases (luckily just local and s3 for now, but that can grow).
To prepare for the future and make things clearer already now, I think we should introduce a new class interface, eg `SavedFile` for files to be saved that have a method for doing the check. In the current case we would have a subclass for local files and another for s3 files. That would allow to keep s3 stuff together in their own class and module, and polymorphism will be used to trigger the correct behaviour.

doc/source/index.rst
```diff
@@ -327,6 +327,27 @@ Options:
   - ``min_valid_data_fraction: 10`` - only generate products if at least 10% of covered
   part of scene contains valid data.
 
+Uploading produced data to S3
```
mraspaud, line 330: I could like to amend this section with a disclaimer that this is NOT the recommended way to dispatch files. The dispatcher should be used instead.

mraspaud (changes requested):
Thanks for all the improvements, but so sorry I still have opinions here...

doc/source/index.rst
```diff
@@ -423,6 +448,7 @@ Example
     - fun: !!python/name:trollflow2.plugins.resample
     - fun: !!python/name:trollflow2.plugins.save_datasets
     - fun: !!python/name:trollflow2.plugins.add_overviews
+    - fun: !!python/name:trollflow2.plugins.s3_uploader
```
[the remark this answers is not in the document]
pnuu: Will do.

doc/source/index.rst
pnuu: Didn't know about the staging zone, can't see it in the documentation. I'll have a look.

trollflow2/launcher.py
pnuu: Which class? The check isn't in any class at the moment.

doc/source/index.rst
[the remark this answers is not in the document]
pnuu: Done in aaa9c65 .

doc/source/index.rst
pnuu: The staging zone is used to rename files after the Satpy writer is done with them, and needs the file to remain in the same filesystem. So it can't be used in this case.

doc/source/index.rst
pnuu: A note added in 3c32b26 .

doc/source/index.rst
pnuu: The functionality of `staging_zone` and file renaming have been modified so that the S3 uploader can use `staging_zone` as temporary local storage.

trollflow2/launcher.py
pnuu: The kwarg is now removed in https://github.com/pytroll/trollflow2/pull/158/commits/42f6639d6887db0df88900c61a1aa3ac932295b6 . The filenames stored in `fmt` now have the eventual file location (as a side-effect of https://github.com/pytroll/trollflow2/pull/158/commits/1e8f88da4268f16c22e8475366e00300a1c5847b ) and `check_results()` can use those URIs directly. So the functionality stayed the same, but the interface is much cleaner.

pnuu:
Ran a test in OpenShift with the latest commits and #159 merged locally and everything is still working.

doc/source/index.rst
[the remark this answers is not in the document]
mraspaud: Thanks! can it be merged with s3_utils?

doc/source/index.rst
```diff
@@ -329,6 +329,32 @@ Options:
...
   part of scene contains valid data.
 
+Uploading produced data to S3
+*****************************
+
+.. note::
+   Outside of container environments, it is perhaps better to have a separate dispatcher from
+   `Trollmoves <https://github.com/pytroll/trollmoves/>`_ to handle the file transfers. With
+   containers this will require an additional container image, configuration and a use of
+   a shared volume for the temporary local files.
```
mraspaud, line 339: I'm not sure I agree here. Nothing prevents containers of running multiple processes simultaneously, as indicated here: https://docs.docker.com/config/containers/multi-service_container/
It is of course not recommended, but fixing a dispatcher as a trollflow2 plugin doesn't change the problem in my opinion.
Moreover, I don't understand why there is a doubt that having product generation and dispatching separate is better. 
Aren't we trying to separate concerns in general?

pnuu: For me this is a placeholder until we can write "directly" to S3. With this approach, we don't need to restructure the processing chain later on by removing the additional dispatcher container and adjusting the configuration of the next step in the chain.

I'll reword the text tomorrow.

doc/source/index.rst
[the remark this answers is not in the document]
pnuu: `trollflow2.s3_utilities.*` has been moved to `trollflow2.plugins.s3` in 8e00ccc

doc/source/index.rst
pnuu: The documentation is reworded and clarified in b13fa0a

trollflow2/launcher.py
```diff
@@ -121,6 +125,23 @@ def check_results(produced_files, start_time, exitcode):
                      f"{elapsed!s}", extra={"time": elapsed})
 
 
+def _check_file(saved_file):
+    file_scheme = urlsplit(saved_file).scheme
+    if file_scheme in ('', 'file'):
+        return _check_local_file(saved_file)
+    if file_scheme == 's3':
+        from trollflow2.plugins.s3 import _check_s3_file
```
mraspaud, line 133: this should be a public method...

doc/source/index.rst
```diff
@@ -329,6 +329,32 @@ Options:
...
+Uploading produced data to S3
+*****************************
+
+.. note::
+   In most cases it is perhaps better to have a separate dispatcher from
+   `Trollmoves <https://github.com/pytroll/trollmoves/>`_ to handle the file transfers. With
+   containers this will require an additional container image, configuration and a use of
+   a shared volume for the temporary local files, so this plugin makes setting up the uploading
+   easier. Later on, if direct saving to S3 becomes available in Satpy writers, this plugin
+   will be removed.
```
mraspaud, line 341: I still don't agree, sorry. I propose we skip the container part and explain what this is: a workaround. Here is a suggestion.

```suggestion
.. note::
   To transfer files to remote file systems, it is recommended to use a separate process, eg the dispatcher 
   from `Trollmoves <https://github.com/pytroll/trollmoves/>`_ to handle the file transfers. However,
   this plugin is a workaround for Satpy's current inability to perform direct remote writing, not a 
   dispatcher in itself. Later on, if direct saving to S3 becomes available in Satpy writers, this plugin
   will be deprecated and eventually removed.
```

trollflow2/tests/test_launcher.py
```diff
@@ -724,59 +725,131 @@ def test_get_dask_client(caplog):
...
+        if skip is None:
+            skip = []
+        self._files = set()
+        for i in range(lo, hi):
+            f = os.path.join(tmp_path, f"file{i:d}")
+            self._files.add(str(f))
+            if i not in skip:
+                if write:
+                    with open(f, mode="wt") as fp:
+                        fp.write("zucchini" * i)
```
mraspaud, line 742: I like zucchinis :)

mraspaud (changes requested):
still some opinions, sorry...

trollflow2/launcher.py
pnuu: The `_check_file()` or `_check_s3_file()`?

doc/source/index.rst
pnuu: Used your suggestion.

trollflow2/launcher.py
mraspaud: `_check_s3_file` since it's from another module

pnuu: Ah, ok. Thought it would be more in-line with `_check_local_file()`. `_check_s3_file()` is changed to public in 83132af

mraspaud (approved):
LGTM, thanks for adding this and bearing with my opinionated reviews! Feel free to merge when the tests pass.

PR #159, opened 2022-09-12
Handle setting subscriber nameserver to False

pnuu:
It wasn't possible to disable the subscriber nameserver to `False`. With this PR, the argument parsing takes care of this. Also, the argument parsing and launching of the `Runner` have now been moved to `trollflow2.launcher` so we could add tests for them.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -23,6 +23,11 @@ The launcher
     :undoc-members:
     :show-inheritance:
 
+It is possible to disable Posttroll Nameserver usage for the incoming
+messages by starting ``satpy_launcher.py`` with command-line arguments
+``-n False -a tcp://<host>:<port>`` where the host and port point to a
+message publisher. Multiple publisher addresses can be given by supplying
+them with additional ``-a`` switches.
 
 Plugins
 -------
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -13,7 +13,7 @@ Welcome to Trollflow2's documentation!
 
 Trollflow2 is an operational generation chain runner for Satpy.
 
-See the example playlist (``pl.yaml``) for inspiration.
+See the bottom of this page and the example playlist (``pl.yaml``) for inspiration.
 
 The launcher
 ------------
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -442,9 +442,9 @@ def sendmail(config, trace):
     pid.terminate()
 
 
-def launch():
+def launch(args_in):
     """Launch the processing."""
-    args = vars(parse_args())
+    args = parse_args(args_in)
 
     log_config = args.pop("log_config", None)
     if log_config is not None:
@@ -466,7 +466,7 @@ def launch():
         runner.run()
 
 
-def parse_args():
+def parse_args(args_in):
     """Parse commandline arguments."""
     parser = argparse.ArgumentParser(
         description='Launch trollflow2 processing with Satpy listening on the specified Posttroll topic(s)')
@@ -493,4 +493,8 @@ def parse_args():
                               "'-a tcp://127.0.0.1:12345 -a tcp://123.456.789.0:9013'"),
                         action="append")
 
-    return parser.parse_args()
+    args = vars(parser.parse_args(args_in))
+    if args['nameserver'].lower() in ('false', 'off', '0'):
+        args['nameserver'] = False
+
+    return args
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -456,7 +456,7 @@ def launch(args_in):
     log_queue = Manager().Queue()
 
     with logging_on(log_queue, log_config):
-        logger.warning("Launching Satpy-based runner.")
+        logger.info("Launching Satpy-based runner.")
         product_list = args.pop("product_list")
         test_message = args.pop("test_message")
         threaded = args.pop("threaded")
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -487,7 +487,7 @@ def parse_args(args_in):
                         help="Log config file (yaml) to use",
                         type=str, required=False)
     parser.add_argument('-n', "--nameserver", required=False, type=str,
-                        help="Nameserver to connect to", default='localhost')
+                        help="Nameserver to connect to. Disable by setting to False", default='localhost')
     parser.add_argument('-a', "--addresses", required=False, type=str,
                         help=("Add direct TCP port connection.  Can be used several times: "
                               "'-a tcp://127.0.0.1:12345 -a tcp://123.456.789.0:9013'"),
--- a/trollflow2/launcher.py
+++ b/trollflow2/launcher.py
@@ -26,7 +26,7 @@ This delegate the actual running of the plugins to a subprocess to avoid any
 memory buildup.
 """
 
-
+import argparse
 import ast
 import copy
 import gc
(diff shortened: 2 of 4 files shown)


trollflow2/launcher.py
```diff
@@ -438,3 +440,61 @@ def sendmail(config, trace):
     pid = Popen([sendmail, "-t", "-oi"], stdin=PIPE)
     pid.communicate(msg.as_bytes())
     pid.terminate()
+
+
+def launch(args_in):
+    """Launch the processing."""
+    args = parse_args(args_in)
```
mraspaud, line 447: do we need `args_in` here, or does argparse get this by itself?

trollflow2/launcher.py
```diff
@@ -438,3 +440,61 @@ def sendmail(config, trace):
...
+    if log_config is not None:
+        with open(log_config) as fd:
+            log_config = yaml.safe_load(fd.read())
+
+    logger = logging.getLogger("satpy_launcher")
+
+    log_queue = Manager().Queue()
+
+    with logging_on(log_queue, log_config):
+        logger.warning("Launching Satpy-based runner.")
```
mraspaud, line 459: Why is this a warning? This would trigger a alarm at our place...

trollflow2/launcher.py
```diff
@@ -438,3 +440,61 @@ def sendmail(config, trace):
...
+        product_list = args.pop("product_list")
+        test_message = args.pop("test_message")
+        threaded = args.pop("threaded")
+        connection_parameters = args
+
+        runner = Runner(product_list, log_queue, connection_parameters, test_message, threaded)
+        runner.run()
+
+
+def parse_args(args_in):
```
mraspaud, line 469: you could make `args_in` optional so that parse_args would get sys.argv by itself I think.

trollflow2/launcher.py
```diff
@@ -438,3 +440,61 @@ def sendmail(config, trace):
     pid = Popen([sendmail, "-t", "-oi"], stdin=PIPE)
     pid.communicate(msg.as_bytes())
     pid.terminate()
+
+
+def launch(args_in):
```
mraspaud, line 445: This function should be tested.

mraspaud (changes requested):
Thanks for the refactor along the way. Some minor things and tests missing, otherwise the logic looks good

trollflow2/launcher.py
```diff
@@ -37,23 +37,25 @@
...
 
-from yaml import UnsafeLoader, SafeLoader, BaseLoader
+from trollflow2.dict_tools import gen_dict_extract, plist_iter
+from trollflow2.logging import logging_on
+from trollflow2.logging import setup_queued_logging
+from trollflow2.plugins import AbortProcessing
 
 
-LOG = getLogger(__name__)
+LOG = logging.getLogger(__name__)
```
gerritholl, line 58: From a coding style point of view, this should probably be `logger` rather than `LOG` (it is not a constant).

trollflow2/launcher.py
gerritholl: Would that be an improvement?  It would be harder to test.

trollflow2/launcher.py
```diff
@@ -438,3 +440,61 @@ def sendmail(config, trace):
...
+                        type=str, required=False)
+    parser.add_argument("-t", "--threaded",
+                        help="Run the product generation in threads instead of processes.",
+                        action='store_true')
+
+    parser.add_argument("-c", "--log-config",
+                        help="Log config file (yaml) to use",
+                        type=str, required=False)
+    parser.add_argument('-n', "--nameserver", required=False, type=str,
+                        help="Nameserver to connect to", default='localhost')
```
gerritholl, line 490: In the main documentation, you write that we can pass `-n False`.  This should be documented here as well.

Would that be for use with a test message?  Otherwise, how would trollflow2 get any input?

gerritholl (changes requested):
Thanks.  I left some comments on how tests and documentation could be improved.

trollflow2/launcher.py
pnuu: For much simpler tests.

trollflow2/launcher.py
pnuu: Oops, fixing to `.info()`.

trollflow2/launcher.py
pnuu: Yeah, I don't want to make testing any more complex than it already is :sweat_smile:

trollflow2/launcher.py
pnuu: Fixed in https://github.com/pytroll/trollflow2/pull/159/commits/7bfb65180c8d5dda8cd6449757d9966edb3cbabc

trollflow2/launcher.py
pnuu: Fixed in https://github.com/pytroll/trollflow2/pull/159/commits/7bfb65180c8d5dda8cd6449757d9966edb3cbabc

trollflow2/launcher.py
pnuu: We use `LOG` throughout the package. I'll create a new issue out of this.

pnuu: https://github.com/pytroll/trollflow2/issues/163

trollflow2/launcher.py
pnuu: Argument parser help message updated in https://github.com/pytroll/trollflow2/pull/159/commits/3c5b38605c995c95a0c90d3e97cf17a79d5b422e

If the `nameserver` is disabled, the messages are received via direct port connection to source defined with `-a` argument as shown in the documentation.

gerritholl (approved):

trollflow2/launcher.py
```diff
@@ -441,3 +443,61 @@ def sendmail(config, trace):
...
+
+
+def launch(args_in):
+    """Launch the processing."""
+    args = parse_args(args_in)
+
+    log_config = args.pop("log_config", None)
+    if log_config is not None:
+        with open(log_config) as fd:
+            log_config = yaml.safe_load(fd.read())
```
mraspaud, line 455: Can we put this in its own function for clarity?

trollflow2/tests/test_launcher.py
```diff
@@ -779,6 +779,28 @@ def qsize(self):
...
+    with mock.patch("trollflow2.launcher.Runner") as Runner:
+        from trollflow2.launcher import launch
+
+        args_in = ['-n', 'localhost', '-a', 'localhost:12345', 'product_list.yaml']
+        launch(args_in)
+        # See that the commandline arguments are passed to Runner
+        assert "'nameserver': 'localhost'" in str(Runner.mock_calls)
+        assert "'addresses': ['localhost:12345']" in str(Runner.mock_calls)
+        assert "product_list.yaml" in str(Runner.mock_calls)
+        Runner.return_value.run.assert_called_once()
```
mraspaud, line 801: Thanks for adding tests!

doc/source/index.rst
```diff
@@ -23,6 +23,11 @@ The launcher
     :undoc-members:
     :show-inheritance:
 
+It is possible to disable Posttroll Nameserver usage for the incoming
+messages by starting ``satpy_launcher.py`` with command-line arguments
+``-n False -a tcp://<host>:<port>`` where the host and port point to a
+message publisher. Multiple publisher addresses can be given by supplying
+them with additional ``-a`` switches.
```
mraspaud, line 30: Just a question here: Is this the same interface we use in other pytroll modules?

mraspaud (commented):
Not much to say, looks pretty good to me :)

trollflow2/launcher.py
pnuu: Done in 95d56a1

doc/source/index.rst
pnuu: Internally, yes. All the packages using `create_subscriber_from_dict_config()` or `create_publisher_from_dict_config()` are using the same settings, explained here: https://github.com/pytroll/posttroll/blob/main/posttroll/subscriber.py#L437-L438

mraspaud: ok! but I was actually wondering if the external/command line interface was the same?

pnuu: For the incoming messages (subscribers) I found these:
- `geographic_gatherer.py` uses `-n false` and `-i tcp://...`, where `-i` is "inbound connection" and means the same as `-a` does here
- `segment_gatherer.py` reads these settings from the config file (using `nameservers` and `addresses` option names)

Internally the both use `posttroll.subscriber.create_subscriber_from_dict_config()`, so given the same combination the `nameserver` handling will be the same. The names used by Posttroll are `addresses` and `nameserver`.

The `-a` and `-n` options were present in `satpy_launcher.py` already before this PR.

mraspaud: ok, we'll keep them for now then

mraspaud (dismissed):
LGTM

PR #168, opened 2022-11-08
Add callback functionality for dask-delayed dataset saving

### LINKED ISSUE #160: Eager processing for faster and more continuous image delivery
## Feature Request

NB: some aspects of this feature request affect multiple Pytroll packages.  I've put this issue here because I think it affects trollflow2 the most.  It's actually two feature requests, but they serve the same goal.

The way we produce images and deliver them to our client software is inefficient.

1. In Pytroll, we wait until all parts of the input data are there (via geographic gatherer or segment gatherer), then send a message to trollflow2 to start processing.  The images would be produced faster if we could start processing each segment as it comes in.
2. In several parts of trollflow2, images are only finalised when all products have been produced.  For example, if we use the file publisher plugin (or any other plugin that comes after `save_datasets`), this is only called after `save_datasets` has completed.  If we use a staging zone, all files are moved from the staging zone to the output directory only when all images have been produced.  When we produce many images, there can be many minutes between the finalisation of the first and the last one.  In this case, it would be beneficial to announce each image (or move it to the output directory) as soon as it is finished.

Both of those aspects cause a delay in the latency of our image production.  This may become more critical with FCI due to the much larger image size compared to SEVIRI.  It also means all our images are delivered to our client software at the same time, which leads to bottlenecks in the images being imported there, that would be avoided or at least reduced if the images were delivered one by one.

I would like a solution that minimises the latency of image delivery to clients without sacrificing quality.  How to do this is of secondary concern.  Problem (1) is difficult to solve if we are doing any resampling.  Problem (2) is in theory easier to solve, but may need significant changes in trollflow2.  Instead of sequentially calling the plugins, it might need to call the filepublisher plugin many times in parallel, while communicating with the save_datasets plugin to know when each file is ready.  Or the save datasets plugin would need to do the publishing itself (or moving the final files in case of using a staging zone or temporary files).

I think either problem should be solvable without breaking existing user workflow, although either problem is large enough that the amount of restructuring needed to solve them may lead to accidental breaking of backwards compatibility.

**Additional context**

For point (1), it would be possible to skip the segment gatherer and send each segment immediately to trollflow2, producing images in native projection, that are then at the end all read and resampled, before stored again, maybe.  I think such a setup could be built without writing new code.  For point (2), we could write directly to the final output directory, but because our file distribution software is external, it doesn't have a way to know when a file is finished (I think).  This leads to the file distribution software copying files before they are finished, in particular for large files that take a while to produce.

gerritholl:
Add functionality for a callback function that is called as soon as a file has been written.  This could be used to ship out written files as soon as they are done, rather than waiting for all files to be complete.

 - [x] Closes #160 partly 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented


gerritholl:
With thanks to @martindurant for the StackOverflow answer at https://stackoverflow.com/a/74354842/974555 pointing out how to do this :)

gerritholl:
Putting on draft as I'm trying to figure out a good way for the callback function to get the information it may need.

gerritholl:
Works wonderfully in the unit tests, but in the real world, the files to be written are empty when the callbacks are called, at least for GeoTIFF.

And the trollflow2 launcher complains about missing files.

gerritholl:
When I collect a bunch of

```
    obj = fake_scene.save_dataset("dragon_top_height", filename=fn, compute=False)
```

and pass them to my wrapper, the files are there and filled.  When I collect a bunch of

```
    obj = trollflow2.plugins.save_dataset(
            {"sargasso": fake_scene},
            {"use_tmp_file": False,
             "fname_pattern": fn,
             "area": "sargasso",
             "product": "dragon_top_height"},
            {"writer": "geotiff"},
            {}, compute=False)
```

and pass them to my wrapper, the files are there but they are empty.

gerritholl:
The difference is that for the NetCDF and simple_image writers, the result of `save_dataset` is a `Delayed` object.  For geotiff, it's a list of a dask array with a RIODataset object.

gerritholl:
I also tried to replace

```
obj = fake_scene.save_dataset("dragon_top_height", filename=fn,
    compute=False)
obs.append(delayed(on_done)(obj, fn))
```

with
```
obj = fake_scene.save_dataset("dragon_top_height", filename=fn,
    compute=False)
    (src, targ) = obj
    objs.append(delayed(on_done)(da.store(src, targ, compute=False), fn))
```

but the problem remains, when `on_done` is called the files still have size 0

gerritholl:
When I add

```python
targ.close()
```

it results in

```
Size for /tmp/test0.tif is 551
Size for /tmp/test2.tif is 551
Size for /tmp/test1.tif is 551
Size for /tmp/test4.tif is 551
Size for /tmp/test3.tif is 551
```

which corresponds to only the header

djhoese:
When you run this code that you tried:

```
obj = fake_scene.save_dataset("dragon_top_height", filename=fn, compute=False)
(src, targ) = obj
objs.append(delayed(on_done)(da.store(src, targ, compute=False), fn))
```

Can you add a print to `on_done` to print out what the argument is? Is it a Delayed object? Or is it `None`? If it is `None` then I'm extremely surprised at the size 0 files. Oh but it sounds like you needed to close the file for it to flush everything to disk, maybe?

gerritholl:
The value of `result` is `(None, None)`:

```
Size for /tmp/test4.tif is 551
[None, None]
Size for /tmp/test3.tif is 551
[None, None]
Size for /tmp/test2.tif is 551
[None, None]
Size for /tmp/test0.tif is 551
[None, None]
Size for /tmp/test1.tif is 551
[None, None]
```

> Oh but it sounds like you needed to close the file for it to flush everything to disk, maybe?

Right.  Example without any pytroll components:

```python
from dask import delayed
import dask.array as da
import os.path
import os
import h5py

ds = da.arange(100).reshape(10, 10)
out = "/tmp/myfile.h5"
os.unlink(out)
f = h5py.File(out, mode='a')

dset = f.create_dataset("data", ds.shape, dtype='f8')

def on_done(result, fn):
    print("Size for", fn, "is now", os.path.getsize(fn))
    return result

x = da.store(ds, dset, compute=False)

da.compute(delayed(on_done)(x, out))
print("Size after computing for", out, "is", os.path.getsize(out))
f.close()
print("Final size for", out, "is", os.path.getsize(out))
```

which results in

```
Size for /tmp/myfile.h5 is now 96
Size after computing for /tmp/myfile.h5 is 96
Final size for /tmp/myfile.h5 is 2848
```

So the issue here appears to be that computing does not result in the file being flushed/closed.  I guess for the other writers, the file gets closed before computation is finished.

We might need to put a `.flush()` somewhere.  Not sure where, and not sure what the impacts of doing so might be.

gerritholl:
I've added a callback that does a close (there's no flush exposed) and it seems to work for GeoTIFF now, at least the unit tests pass.  Will do a live test later today.

gerritholl:
Passes unit tests, but fails the field test with

```
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/trollflow2/launcher.py", line 377, in process
    cwrk.pop('fun')(job, **cwrk)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/trollflow2/plugins/__init__.py", line 334, in save_datasets
    compute_writer_results(objs)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/satpy/writers/__init__.py", line 536, in compute_writer_results
    da.compute(delayeds)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/dask/base.py", line 602, in compute
    results = schedule(dsk, keys, **kwargs)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/dask/threaded.py", line 89, in get
    results = get_async(
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/dask/local.py", line 511, in get_async
    raise_exception(exc, tb)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/dask/local.py", line 319, in reraise
    raise exc
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/dask/local.py", line 224, in execute_task
    result = _execute_task(task, data)
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/dask/core.py", line 119, in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
  File "/opt/pytroll/pytroll_inst/mambaforge/envs/pytroll-py310/lib/python3.10/site-packages/trollflow2/plugins/__init__.py", line 974, in callback_close
    obj[1].close()
AttributeError: 'tuple' object has no attribute 'close'
```

inspection reveals that `obj` here is:

```
[(array([[[ 27,  27,  20, ...,  29,  29,  29],
        [ 27,  27,  20, ...,  28,  28,  28],
        [ 27,  27,  20, ...,  28,  28,  28],
        ...,
        [ 24,  23,  23, ..., 103, 103, 100],
        [ 23,  23,  23, ..., 103, 103, 100],
        [ 23,  23,  23, ..., 100,  97,  97]]], dtype=uint8), 6, 119), (<trollimage.xrimage.RIODataset object at 0x7effbea3d480>, <trollimage.xrimage.RIOTag object at 0x7effbea3eef0>, <trollimage.xrimage.RIOTag object at 0x7eff7109c910>)]
```

meaning that rather than target being a single `RIODataset`, it is actually a tuple of `(RIODataset, RIOTag, RIOTag)` objects…

gerritholl:
A quick test shows that for this case, everything seems to work as expected if I loop through `obj[1]` and close all members, but why is `obj[1]` sometimes a `RIODataset` and sometimes a tuple?

djhoese:
I mentioned the reasoning for the multi-element target on slack (tiff tags using dask arrays) and that we should maybe add a wrapper object in XRImage. For the `on_done` callback, should it maybe be doing the closing? Or maybe be chained after another function that does the closing?

gerritholl:
It seems to pass a system test now, with one caveat: when we use the `callback_move` plugin, the launcher complains about missing files.

gerritholl:
The mover callback could use `staging_zone`, so that we can have a 1-1 mapping between temporary path and final path.  Then, in `save_datasets`, In `job['produced_files'].put(fmat_config['filename'])`, we could already put the post-moving filename such that this is where `check_datasets` will look later.

gerritholl:
Converting back to draft because I'm still getting empty/black/tiny files being produced, despite passing unit tests… I thought it was working when I tested it earlier…

gerritholl:
If I pass

`delayed(testlog)(da.store(src, targ, compute=False), ...)`

then it seems to work, but so far I am passing

`delayed(testlog)(scn.save_dataset(..., compute=False), ...)`

and then it doesn't work, in fact the file remains full of zeroes even at the end, even if the result of `delayed(testlog)` is passed to `compute_writer_results`.

gerritholl:
That means that we have to make changes to `plugins.save_datasets` in trollflow2 and/or `compute_writer_results` in Satpy.

gerritholl:
I wonder what the performance different would be between:

```python
da.compute(da.store(sources, targets, compute=False))
```

and

```python
da.compute([da.store(src, targ, compute=False) for (src, targ) in zip(sources, targets)])
```

We may have to do the latter for early moving to work for GeoTIFF.

gerritholl:
> I mentioned the reasoning for the multi-element target on slack (tiff tags using dask arrays) and that we should maybe add a wrapper object in XRImage.

It's not only tiff tags using dask arrays.  It's also multiple files.  We get

```[<trollimage._xrimage_rasterio.RIODataset object at 0x7fc43ea23bb0>, <trollimage._xrimage_rasterio.RIODataset object at 0x7fc43ebab010>, <trollimage._xrimage_rasterio.RIODataset object at 0x7fc43ea47d90>]```

Although I could close all three of them, the whole point of the exercise is that I would move _each_ file as soon as it's finished.  A `callback_move` would have to be called three times, once for each file.  That's not how Satpy's `compute_writer_results` calls `da.store`.

**Comment by djhoese** [comment]
Ah ok, so the writer returns multiple files, but XRImage only deals with one file at a time.

**Comment by gerritholl** [comment]
Needs https://github.com/pytroll/satpy/pull/2281

**Comment by gerritholl** [comment]
It seems to work now with https://github.com/pytroll/satpy/pull/2281, but there appears to be a problem with resource usage.  There is a large increase in RAM usage and full disc processing that previously used 80 GB is running out of RAM on a 128 GB machine.

**Comment by gerritholl** [comment]
Calculation for FCI over Europe without callbacks:

tf2-dask-without-callbacks

with callbacks:

tf2-dask-with-callbacks

**Comment by gerritholl** [comment]
In a Satpy context, I don't find much of a performance impact, see https://github.com/pytroll/satpy/pull/2281#issuecomment-1319003086.  But when I make a comparison through trollflow2, I see major differences.  MCVE:

```python
import os
import copy
from unittest import mock
from dask.diagnostics import Profiler, ResourceProfiler, visualize
from sattools.io import plotdir
from glob import glob
from satpy import Scene

input_mda = {}

job = {}
job['input_mda'] = input_mda

sc = Scene(filenames={"seviri_l1b_hrit": seviri_files})
names = sc.available_dataset_names()
sc.load(names)

def noop(obj, src, targ, job, fmat_config):
    """Do nothing."""
    return obj

form = [
    {"writer": "geotiff", "fill_value": 0},
]

#call_on_done = None
call_on_done = [noop]
product_list = {
    "fname_pattern": "{productname}-{writer}.tif",
    "output_dir": os.fspath(plotdir()),
    "areas": {
        "sargasso": {
            "products": {
                nm: {
                    "productname": nm,
                    "formats": copy.deepcopy(form)}
                for nm in names
                },
            }
        }
    }
if call_on_done:
    product_list["call_on_done"] = call_on_done
job['resampled_scenes'] = {"sargasso": sc}
job['produced_files'] = mock.MagicMock()

if call_on_done:
    d = "with"
else:
    d = "without"
with Profiler() as prof, ResourceProfiler(dt=0.05) as rprof:
    save_datasets(job)
visualize([prof, rprof], show=False, save=True, filename=os.fspath(plotdir() /
    f"dask-profile-{d:s}-call-on-done.html"))
```

Without the wrapper, it takes 15.1 seconds and uses 1.4 GB RAM.

image


With the wrapper, it takes 18.5 seconds and uses 2.5 GB RAM.

image

the dask graphs look strange, with the profiler showing 8 tasks in parallel at a time that the resource profiles shows 100% CPU, and the profiler shows nothing happening after 3 or 6 seconds even if the resource profiler shows there is still ongoing work...

djhoese:
It looks like you are chaining the delayed objects being computed. Is this necessary? Or can you do `delayeds.extend(` call and have the delayed handling return a list of delayed objects (`[callback(delayed, ...) for callback in callbacks]`)? I'm wondering if the chaining is making it so the result of one Delayed function is being held on to because it is "needed" by the next one and so on.

gerritholl:
The dask graph looks very different between without and with the callback, and the SVG increases in size from 6.9M to 8.3M.  Profiling also shows that `_compute_quantile` (in xarray) gets called 8 times without and 16 times with the callback, even though it only appears 8 times in the dask graph.

gerritholl:
> Is this necessary? 

> I'm wondering if the chaining is making it so the result of one Delayed function is being held on to because it is "needed" by the next one and so on.

I think this is necessary.  The first wrapper is closing.  The closing must be completed before the moving, because closing flushes data to disk so only after closing I know that the writing is complete.

gerritholl:
But the chaining cannot explain the problem I observe here, because in this test I am only applying a single callback.

gerritholl:
dask graph with noop-wrapper:

dask-delayed-tf2-test

dask graph without noop-wrapper:

dask-delayed-tf2-test-without-wrapper

I cannot (yet) reproduce this big difference in the dask graph outside of my trollflow2 changes, so it may be due to something I've done wrongly.

gerritholl:
Not passing `src` only partly solves the problem.  I no longer get big increases in RAM for large productions.  However, I get an increase in RAM for small productions and an increase in run-time for large productions.

Taking the full FCI test dataset to produce 16 channels and 7 RGBs, all resampled to either a 1 km or a 2 km grid covering Germany, Europe (roughly LAC4) or the full disc.  Peak RAM and wall clock time usage without or with callbacks that close the file, move it, and write a log message:

| Area | RAM without | time without | RAM with | time with |
| ------------- | ------------- | ----- | ----- | ----- |
| Germany | 2.2 GB | 0:24.2 | 3.0 GB | 0:24.7
| Europe | 20.0 GB | 02:06.7 | 20.4 GB | 02:40.9
| Full disc | 84.0 GB | 08:39.4 | 65.9 GB | 11:32.3

- For a small area the RAM increases and the time stays the same.
- For a medium sized area the RAM stays the same and the time increases.
- For a large area the RAM decreases and the time increases.

gerritholl:
When I use the synchronous scheduler for debugging purposes, producing FCI over Europe (LAC4) without enhancing or resampling is *faster* with callbacks than without.  It's slightly faster when the callback does nothing, but it's *much* faster when the callback is closing files.

| What | RAM, baseline | time, baseline | RAM, close/log | time, close/log | RAM, noop | time, noop
| ------------- | ------------- | ----- | ----- | ----- | ----- | ----- |
| default scheduler | 3.4 GB | 0:48 | 3.7 GB | 1:02 | 3.3 GB | 1:13 | 
| synchronous scheduler | 2.9 GB | 2:11 | 2.8 GB | 1:22 | 2.8 GB | 2:00

gerritholl:
Tests are failing because it depends on merged but unreleased satpy capabilities.

pnuu:
I guess it makes sense that for large area the RAM usage is lower when the fully used data can be released and cleaned.

trollflow2/plugins/__init__.py
```diff
@@ -286,6 +291,28 @@ def save_datasets(job):
...
+    could be used, for example, to ship products as soon as they are
+    successfully produced.  Three callback functions are provided
+    with trollflow2: :func:`callback_log`, :func:`callback_move`, and
+    :func:`callback_close`.  If using the geotiff or ninjogeotiff writers,
+    :func:`callback_close` should be used before the others for correct
+    results.  When using :func:`callback_move`, the user must also set
+    ``early_moving`` to True and use a ``staging_zone``.  If both are
+    used, :func:`callback_log` must be called AFTER :func:`callback_move`,
+    because :func:`callback_log` searches for the final destination of
+    the file and reports the size (so it accesses the metadata).
```
pnuu, line 314: This could be a separate chapter.

Should the functionality of these built-in callback functions explained here? Now the docstring just says how they should be used, not what they do.

trollflow2/plugins/__init__.py
```diff
@@ -296,16 +323,106 @@ def save_datasets(job):
...
+        callbacks = [dask.delayed(c) for c in call_on_done]
+    else:
+        callbacks = None
+    with renamed_files(not early_moving) as renames:
         for fmat, fmat_config in plist_iter(job['product_list']['product_list'], base_config):
-            obj = save_dataset(scns, fmat, fmat_config, renames, compute=eager_writing)
-            if obj is not None:
-                objs.append(obj)
+            late_saver = save_dataset(scns, fmat, fmat_config, renames, compute=eager_writing)
+            late_saver = _maybe_apply_callbacks(late_saver, callbacks, job, fmat_config)
```
pnuu, line 336: Sounds like it's random whether the callbacks are called or not :sweat_smile: I think we've named this kind of conditionals simply `_apply_foo()` or such in other packages.

trollflow2/plugins/__init__.py
```diff
@@ -296,16 +323,106 @@ def save_datasets(job):
...
+
+    If we are using callbacks via the ``call_on_done`` parameter, wrap
+    ``late_saver`` with those iteratively.  If not, return ``late_saver`` as is.
+    Here, ``late_saver`` is whatever :meth:`satpy.Scene.save_datasets`
+    returns.
+    """
+    if callbacks is None:
+        return late_saver
+    if isinstance(late_saver, Delayed):
+        return _apply_callbacks_to_delayed(late_saver, callbacks, None, None, *args)
```
pnuu, line 355: Should there be a test for a case where the writer is returning a `Delayed` object?

trollflow2/plugins/__init__.py
pnuu: Looking at the callback docstrings I think it's even better if there's just a (bullet) list here that links to the corresponding docstrings. The individual docstrings are easier at least for me to understand than this longer version. So something like
```
Three built-in are provided with Trollflow2: :func:`callback_close`, :func:`callback_move` and :func:`callback_log`.
```

pnuu (commented):
Just couple of comments, otherwise LGTM.

mraspaud:
@gerritholl I made an attempt at fixing the merge conflicts

pnuu:
I fixed the two logger names that were not present in my logging update already merged in `main`.

trollflow2/plugins/__init__.py
gerritholl: Yes, there should.  And there was a bug for this case too :)  Added and fixed in https://github.com/pytroll/trollflow2/pull/168/commits/1f56dd1f6f279aec198118f7cb3d4f0a02aadebf

trollflow2/plugins/__init__.py
gerritholl: Renamed to `_apply_callbacks` in https://github.com/pytroll/trollflow2/pull/168/commits/1f56dd1f6f279aec198118f7cb3d4f0a02aadebf

pnuu (approved):
LGTM

trollflow2/plugins/__init__.py
```diff
@@ -302,16 +324,106 @@ def save_datasets(job):
     base_config = job['input_mda'].copy()
     base_config.pop('dataset', None)
     eager_writing = job['product_list']['product_list'].get("eager_writing", False)
-    with renamed_files() as renames:
+    early_moving = job['product_list']['product_list'].get("early_moving", False)
+    sentinel = object()
+    call_on_done = job["product_list"]["product_list"].get("call_on_done", sentinel)
+    if call_on_done is not sentinel:
+        callbacks = [dask.delayed(c) for c in call_on_done]
```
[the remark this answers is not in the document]
gerritholl: It could.  I don't feel very strongly either way.  My reason: accepting None makes the API slightly more complicated, as we then accept two very different argument types (sequence or None), which would need to be marked wherever we document valid values for `call_on_done`, as well as in tests.  By defining a sentinel inside the function, passing None is invalid and needs to be neither documented nor tested.

trollflow2/plugins/__init__.py
```diff
@@ -260,16 +264,17 @@ def _create_data_query(product, res):
...
     """Context renaming files."""
     renames = {}
 
     yield renames
 
-    for tmp_name, actual_name in renames.items():
-        target_scheme = urlsplit(actual_name).scheme
-        if target_scheme in ('', 'file'):
-            os.rename(tmp_name, actual_name)
+    if do_renames:
```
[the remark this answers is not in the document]
gerritholl: Finally I found another solution using `nullcontext()`, that I implemented in 8a37c43.

pnuu:
I fixed the couple merge conflicts. At least I hope I did.

trollflow2/plugins/__init__.py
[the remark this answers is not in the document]
gerritholl: Changed to using `None` in 6bd0218

mraspaud (approved):
LGTM, thanks for adding this feature!

PR #173, opened 2022-12-14
Make it possible to disable nameserver in FilePublisher

pnuu:
This PR switches the `FilePublisher` plugin to use `create_publisher_from_dict_config()` from Posttroll.

The undocumented feature of turning `nameserver` usage off uses now the same principles that Posttroll:
- use `nameservers: False`
- set `port` to non-zero numeric value

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented


mraspaud (approved):
LGTM. I'm not so glad that we need to mock so much, but I think that's a problem in posttroll, where we should provide a testing context...

pnuu:
Thanks! Merging.

PR #175, opened 2023-03-08
Drop Python 3.8 support

pnuu:
This PR drops Python 3.8 support from Trollflow2.

The CI is adjusted to test with Python 3.9, 3.10 and 3.11.

 - [ ] Tests passed

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,10 +10,10 @@ jobs:
       fail-fast: true
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.8", "3.9", "3.10"]
+        python-version: ["3.9", "3.10", "3.11"]
         experimental: [false]
         include:
-          - python-version: "3.10"
+          - python-version: "3.11"
             os: "ubuntu-latest"
             experimental: true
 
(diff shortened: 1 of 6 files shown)


gerritholl (dismissed):
Looks good to me.

pnuu:
Should we also remove OSX from the tests? They seem to be stuck to "expected" also in other PRs.

gerritholl:
Do we have any OSX users?  If not, I think we can remove the tests, and perhaps document somewhere that OSX is not supported.

pnuu:
I think we've already dropped OSX tests from CI. At least it's not listed in `ci.yaml`. I'll add some documentation.

pnuu:
OSX tests were removed in 27b6de405a58f1728804f0eecb6ba8d95b50457f

mraspaud (approved):
LGTM

mraspaud:
I don't understand why macos action are still being triggered :thinking:

pnuu:
Looking at https://github.com/pytroll/trollflow2/actions/runs/4362136850/workflow the OSX tests are not triggered, but for some reason still reported on the PR page. Weird.

mraspaud (approved):
LGTM

PR #191, opened 2023-10-02
Fix call to posttroll publishers' start and stop

mraspaud:
This PR fixes the calls to publisher methods to be inline with the newer version of posttroll, and avoid a hang at the end of the processing.

 - [x] Tests added  
 - [x] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,10 +36,10 @@ except ImportError:
     pass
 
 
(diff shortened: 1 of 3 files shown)


pnuu (approved):
LGTM, thanks for the fix!

pnuu (approved):
LGTM

PR #184, opened 2023-06-02
Add a Command Line Interface

mraspaud:
This PR adds a command line interface (CLI) to trollflow2.

This allows for easier launching in environments that might not have posttroll messaging enabled or for batch reprocessing.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Passes ``flake8 trollflow2`` 
 - [ ] Fully documented 
 - [ ] Add your name to `AUTHORS.md` if not there already

PR #193, opened 2023-10-24
Profile CLI run through dask

gerritholl:
For the CLI added in #184, add an option to profile the results using dask.diagnostics and visualize the results as a bokeh plot.

 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented

### DIFF
--- a/trollflow2/cli.py
+++ b/trollflow2/cli.py
@@ -1,10 +1,12 @@
 """Trollflow2 command line interface."""
 
 import argparse
+import contextlib
 import json
 import logging
 from queue import Queue
 
+import dask.diagnostics
 import yaml
 
 from trollflow2.launcher import logging_on, process_files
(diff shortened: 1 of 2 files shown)


gerritholl:
We should merge #184 first.

djhoese:
In polar2grid I think I called this --create-profile and nargs=?. That way I could provide a default HTML file name with info about the reader and writer being used and the current time.

gerritholl:
> In polar2grid I think I called this --create-profile and nargs=?. That way I could provide a default HTML file name with info about the reader and writer being used and the current time.

I don't understand what you mean by "provide a default HTML file name".  With `nargs=?` you'd be giving multiple HTML file names to write to?  Info about the reader and writer, you mean by string substitution?

djhoese:
Sorry. With `nargs="?"` and `default=False` you can pass only `--create-profile` to the CLI and `args.create_profile` will be `None` (if I remember correctly). Or you can do `--create-profile profile_my_filename.html`. Or if it isn't specified at all then `args.create_profile` is `False`. If it is `None` I then later set it to `"{project_name}_{glue_name}_{start_time:%Y%m%d_%H%M%S}.html"` where for P2G `project_name` is either "geo2grid" or "polar2grid", `glue_name` is `<reader_name>2<writer_name>` and start_time is the current time when the script is executed. I also allow the user to specify `end_time` in the format string.

I turned it into a context manager:

https://github.com/ssec/polar2grid/blob/4c470fa9de54c4b141cf4a5301a90005f35d5c84/polar2grid/glue.py#L241-L266

You'll notice I also print out a `file://` version of the HTML path so in my Terminal on linux it is Ctrl + left click-able and opens in my browser.

pnuu:
I merged `main` to the PR and tried to resolve the conflicts. Lets see what the CI says.

pnuu (approved):
LGTM

mraspaud (approved):
LGTM

PR #199, opened 2024-02-12
Fix logging to handler config without loggers

mraspaud:
This PR allows trollflow2 to accept log configs without loggers.

 - [x] Closes #197  
 - [x] Closes #192 
 - [x] Closes #183 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Passes ``flake8 trollflow2``

PR #201, opened 2024-02-28
Fix test failures

pnuu:
This PR fixes couple of test failures.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented 

### DIFF
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -1552,9 +1552,9 @@ class TestCheckMetadata(TestCase):
         with mock.patch('trollflow2.plugins.get_config_value') as get_config_value:
             get_config_value.return_value = None
             job = {'product_list': None, 'input_mda': {'start_time': dt.datetime(2020, 3, 18)}}
-            self.assertIsNone(check_metadata(job))
-            get_config_value.return_value = {'start_time': -2e6}
-            self.assertIsNone(check_metadata(job))
+            assert check_metadata(job) is None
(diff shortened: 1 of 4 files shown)


trollflow2/tests/test_trollflow2.py
```diff
@@ -1554,7 +1554,8 @@ def test_discard_old_data(self):
             job = {'product_list': None, 'input_mda': {'start_time': dt.datetime(2020, 3, 18)}}
             self.assertIsNone(check_metadata(job))
             get_config_value.return_value = {'start_time': -2e6}
-            self.assertIsNone(check_metadata(job))
+            with self.assertRaises(AbortProcessing):
+                check_metadata(job)
```
mraspaud, line 1558: I don't think that's the right solution, and I actually already fixed this in https://github.com/pytroll/trollflow2/pull/199/commits/99d4a6b9a6bbad8f04dc7874a56b393c0eea5dbf
(https://github.com/pytroll/trollflow2/pull/199 waited for a positive review to be auto merged, but realise it was only a comment. merging now)

mraspaud (changes requested):
Looks good, but the time test needs to be fixed (see inline comment).

trollflow2/tests/test_trollflow2.py
pnuu: Fixed by rebasing with current `main`.

mraspaud (dismissed):
LGTM

mraspaud (approved):
LGTM

PR #205, opened 2024-06-13
Add plugins for fsspec caching and cache cleaning

pnuu:
This PR adds the possibility to do caching with `fsspec`. Mostly useful for remotely read data that has a complicated structure which slows down the reading such as FCI L1c in S3 storage. Additional plugin is available for cleaning the cache.

 - [x] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -1138,3 +1138,12 @@ def use_fsspec_cache(job):
         open_files = fsspec.open_files(cached_filenames)
     fs_files = [FSFile(open_file) for open_file in open_files]
     job["input_filenames"] = fs_files
+
+
+def clear_fsspec_cache(job):
+    """Clear all files in fsspec cache directory."""
+    filenames = job["input_filenames"]
+
+    for f in filenames:
+        if hasattr(f, "_fs"):
+            f._fs.clear_cache()
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -1114,3 +1114,15 @@ def callback_close(obj, targs, job, fmat_config):
         for targ in targs:
             targ.close()
     return obj
+
+
+def use_fsspec_cache(job):
+    """Use the caching from fsspec for (remote) files."""
+    import fsspec
+    from satpy.readers import FSFile
+
+    cache = job["product_list"]["fsspec_cache"]
+    filenames = job["input_filenames"]
+    open_files = fsspec.open_files([f"{cache}::{f}" for f in filenames])
+    fs_files = [FSFile(open_file) for open_file in open_files]
+    job["input_filenames"] = fs_files
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -1122,7 +1122,19 @@ def use_fsspec_cache(job):
     from satpy.readers import FSFile
 
     cache = job["product_list"]["fsspec_cache"]
+    cache_dir = job["product_list"].get("fsspec_cache_dir")
     filenames = job["input_filenames"]
-    open_files = fsspec.open_files([f"{cache}::{f}" for f in filenames])
+    cached_filenames = [f"{cache}::{f}" for f in filenames]
+    if cache == "blockcache":
+        open_files = fsspec.open_files(cached_filenames,
+                                       blockcache={"cache_storage": cache_dir})
+    elif cache == "filecache":
+        open_files = fsspec.open_files(cached_filenames,
+                                       filecache={"cache_storage": cache_dir})
+    elif cache == "simplecache":
+        open_files = fsspec.open_files(cached_filenames,
+                                       simplecache={"cache_storage": cache_dir})
+    else:
+        open_files = fsspec.open_files(cached_filenames)
     fs_files = [FSFile(open_file) for open_file in open_files]
     job["input_filenames"] = fs_files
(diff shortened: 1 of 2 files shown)


pnuu:
The failing test is due to a small change in Posttroll and the way we've implemented the tests (with `mock`). This shows up only in unstable tests because the change is not published.

doc/source/plugins.rst
```diff
@@ -359,3 +359,20 @@ Settings:
 The S3 connection options are handled by the
 `fsspec <https://filesystem-spec.readthedocs.io/en/latest/features.html#configuration>`_
 configuration mechanism.
+
+Caching remotely read data
+**************************
+
+The ``use_fsspec_cache`` plugin can be used to turn on cachcing for remotely read files. See below for
```
mraspaud, line 366:
```suggestion
The ``use_fsspec_cache`` plugin can be used to turn on caching for remotely read files. See below for
```
typo

trollflow2/tests/test_trollflow2.py
```diff
@@ -2018,7 +2019,8 @@ def test_filepublisher_kwargs_direct_instance_no_nameserver(self):
 
             _ = FilePublisher(port=40000, nameservers=False)
             NoisyPublisher.assert_not_called()
-            Publisher.assert_called_once_with('tcp://*:40000', name='l2processor', min_port=None, max_port=None)
+            assert "tcp://*:40000" in Publisher.mock_calls[0].args
```
mraspaud, line 2022: the uniqueness check of the call is dropped, is this on purpose?

doc/source/plugins.rst
```diff
@@ -359,3 +359,20 @@ Settings:
...
+cache cleaning.
+
+Settings used from the product list:
+  - ``fsspec_cache`` - name of the cache implementation. Can be one of ``blockcache``, ``filecache`` or
+    ``simplecache``
+  - ``fsspec_cache_dir`` - location where the caching is done. If it does not exist, it will be created
```
mraspaud, line 372: I think it would be nicer to have the two config options under the same section, and maybe also a way to pass any args to the cache, eg:
```yaml
fsspec_cache:
  type: filecache
  cache_args: <fsspec cache arguments>
```
and refer to the fsspec cache doc

trollflow2/plugins/__init__.py
```diff
@@ -1114,3 +1114,37 @@ def callback_close(obj, targs, job, fmat_config):
...
+    if cache_dir:
+        if cache == "blockcache":
+            open_files = fsspec.open_files(cached_filenames,
+                                           blockcache={"cache_storage": cache_dir})
+        elif cache == "filecache":
+            open_files = fsspec.open_files(cached_filenames,
+                                           filecache={"cache_storage": cache_dir})
+        elif cache == "simplecache":
+            open_files = fsspec.open_files(cached_filenames,
+                                           simplecache={"cache_storage": cache_dir})
```
mraspaud, line 1137: how about
```python
kwargs = {cache: {"cache_storage":...}}
open_files = fsspec.open_files(cached_filenames, **kwargs)
```

mraspaud (commented):
Nice work! I'm wondering if this should really be its own plugin, or just an option to the reading plugin?

trollflow2/tests/test_trollflow2.py
pnuu: Yes. The test result depended on the Posttroll version (with current `main` the `min_port` and `max_port` are not in the default call) so this is more generic.

mraspaud: ok, but should we check that it is only called once?

pnuu: Oh, yeah. Will do that.

trollflow2/plugins/__init__.py
pnuu: Fixed in https://github.com/pytroll/trollflow2/pull/205/commits/f2701dbf0119780768b780fa721222b166f14d35

doc/source/plugins.rst
pnuu: Fixed in https://github.com/pytroll/trollflow2/pull/205/commits/1c2844e9249f375897490a1f85d8540413327c1c

trollflow2/tests/test_trollflow2.py
pnuu: Fixed in https://github.com/pytroll/trollflow2/pull/205/commits/76de724eb4fe201653a40836dee4b279b73b1c5d

trollflow2/plugins/__init__.py
```diff
@@ -1114,3 +1114,29 @@ def callback_close(obj, targs, job, fmat_config):
...
+    job["input_filenames"] = fs_files
+
+
+def clear_fsspec_cache(job):
+    """Clear all files in fsspec cache directory."""
+    filenames = job["input_filenames"]
+
+    for f in filenames:
+        if hasattr(f, "_fs"):
+            f._fs.clear_cache()
```
mraspaud, line 1142: It's not so nice that we need to access a private attribute of FSFile, should we add a public `fs` property (read-only) to FSFile instead?

mraspaud (commented):
About the _fs...

trollflow2/plugins/__init__.py
pnuu: Yeah, that would be great. I'll have a go tomorrow if I remember.

pnuu: Did this already today: https://github.com/pytroll/satpy/pull/2832

pnuu: I've now updated this PR to use the public `.fs` property to clean the cache. The tests will fail until a new Satpy release is available.

mraspaud (approved):
LGTM!

PR #208, opened 2024-07-30
Move import to be dask-dist friendly

gerritholl:
Make trollflow2 more multiprocessing / dask-distributed friendly. Move the launcher import in satpy_launcher.py within the main function, so that it only gets executed in the primary process. The call to `Manager()` in `trollflow2.__init__.py` was causing attempts to use trollflow2 with dask distributed to fail with a RuntimeError.

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented 

I don't know how to add a unit test for this.


pnuu (approved):
LGTM

PR #210, opened 2024-08-23
Added crop functionality to trollflow2

Vpirm9:
An addition for using the scene.crop() functionality from `satpy` has been added to `trollflow2. It can be called in the `trollflow2.yaml` file with `- fun: !!python/name:trollflow2.plugins.crop. This functionality was implemented because I was tasked with cropping geostationary data and saving it to the local archive. The aim of this approach was to preserve the data in its original form while reducing overall storage space by retaining only the data over the region of interest.

The `crop()` function was added, which uses the same output `job["resampled_scenes"]` as `resample(), so they cannot be used simultaneously. However, no situations have been identified where using both functions at the same time would be necessary.

 - [x] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented 

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -170,6 +170,25 @@ def resample(job):
             logger.debug("area: %s, area_conf: %s", area, str(area_conf))
             job['resampled_scenes'][area] = scn.resample(area, **area_conf)
 
+def crop(job):
+    """Crop the scene to some areas. Cropping to the llbox and xybox are not supported jet."""
+
+    product_list = job["product_list"]
+    conf = _get_plugin_conf(product_list, "/product_list", "")
+    job["resampled_scenes"] = {}
+    scn = job["scene"]
+    for area in product_list["product_list"]["areas"]:
+        area_conf = _get_plugin_conf(
+            product_list, "/product_list/areas/" + str(area), conf
+        )
+        logger.info("Cropping to %s", str(area))
+
+        # Load composites before cropping, otherwise cropping fails for composites
+        scn.load(scn.wishlist, generate=True)
+        # If area is none, no cropping is done.
+        if area != "None":
+            logger.debug("area: %s, area_conf: %s", area, str(area_conf))
+            job["resampled_scenes"][area] = scn.crop(area)
 
 # Datasets saving
 


pnuu (changes requested):
Could you add some unit tests for the new feature? Also add documentation for the new plugin to `doc/source/plugins.rst`.

trollflow2/plugins/__init__.py
```diff
@@ -170,6 +170,25 @@ def resample(job):
...
+    job["resampled_scenes"] = {}
+    scn = job["scene"]
+    for area in product_list["product_list"]["areas"]:
+        area_conf = _get_plugin_conf(
+            product_list, "/product_list/areas/" + str(area), conf
+        )
+        logger.info("Cropping to %s", str(area))
+
+        # Load composites before cropping, otherwise cropping fails for composites
+        scn.load(scn.wishlist, generate=True)
```
Manny7717, line 187: This runs on every loop iteration — composites get (re)loaded per area. The resample plugin above guards with `if not set(scn.keys()).issuperset(scn.wishlist)`. Hoist above the loop or reuse that guard.

trollflow2/plugins/__init__.py
```diff
@@ -170,6 +170,25 @@ def resample(job):
...
+        # If area is none, no cropping is done.
+        if area != "None":
```
Manny7717, line 189: String 'None' as a skip sentinel matches resample's convention (consistent!), but any area literally named 'None' in config is silently skipped. A brief comment or debug log would help future users.

trollflow2/plugins/__init__.py
```diff
@@ -170,6 +170,25 @@ def resample(job):
             logger.debug("area: %s, area_conf: %s", area, str(area_conf))
             job['resampled_scenes'][area] = scn.resample(area, **area_conf)
 
+def crop(job):
+    """Crop the scene to some areas. Cropping to the llbox and xybox are not supported jet."""
```
Manny7717, line 174: Typo: "not supported jet" -> "not supported yet"

Manny7717 (commented):
## Review

Useful plugin to have — partial-disk products are a real gap in the resample pipeline. A few things before this is ready to merge:

1. **Composites are loaded inside the area loop** (inline at line 187) — `scn.load(scn.wishlist, generate=True)` re-runs for every area. The `resample` plugin guards this with `if not set(scn.keys()).issuperset(scn.wishlist):` so loading only happens once (and only when needed); mirror that pattern, or hoist the load above the loop.
2. **The `'None'` sentinel** (inline at line 189) matches the `resample` plugin's convention, so it's consistent — but that also means an area literally named `None` in the product config is silently skipped. Worth a comment or a `logger.debug` explaining the sentinel.
3. **`job["resampled_scenes"] = {}` clobbers** whatever `resample` may have already put there if both plugins are chained in one product list. If `crop` is intended as a resample *alternative*, consider documenting that (or preserving existing entries).
4. **No tests added** — trollflow2 has a tests/ directory with plugin-adjacent tests; a test for `crop` with a mocked scene (assert load is called once and `scn.crop` per area) would lock in the behavior.
5. Docstring typo: "not supported jet" -> "yet" (inline at 174).

Question on config shape: `resample` accepts proj-dicts from the product config, but satpy's `Scene.crop` takes an `AreaDefinition` or an ll/xy bbox dict — not a proj dict. What does the `areas` config carry in the crop case? If it's an AreaDefinition from a registered area, fine; if it's the same proj-dicts resample consumes, `scn.crop(area)` will raise.

PR #217, opened 2024-10-17
Add handling for SIGTERM

pnuu:
This PR adds handling for SIGTERM signal, so that Trollflow2 can be terminated gracefully in for example container environment.

 - [ ] Closes #xxxx 
 - [x] Tests added  
 - [x] Tests passed 
 - [x] Fully documented 

### DIFF
--- a/doc/source/launcher.rst
+++ b/doc/source/launcher.rst
@@ -11,3 +11,7 @@ messages by starting ``satpy_launcher.py`` with command-line arguments
 ``-n False -a tcp://<host>:<port>`` where the host and port point to a
 message publisher. Multiple publisher addresses can be given by supplying
 them with additional ``-a`` switches.
+
+The Trollflow2 process can be terminated gracefully by sending SIGTERM
+to the it, eg. by ``kill -SIGTERM <pid>`` where ``<pid>`` is the
+process ID number of Trollflow2.
--- a/doc/source/launcher.rst
+++ b/doc/source/launcher.rst
@@ -14,4 +14,6 @@ them with additional ``-a`` switches.
 
 The Trollflow2 process can be terminated gracefully by sending SIGTERM
 to the it, eg. by ``kill -SIGTERM <pid>`` where ``<pid>`` is the
-process ID number of Trollflow2.
+process ID number of Trollflow2.  After this no new messages will be
+accepted, and the completion of currently running processing will exit
+Trollflow2.
--- a/trollflow2/tests/test_launcher.py
+++ b/trollflow2/tests/test_launcher.py
@@ -887,15 +887,15 @@ def test_generate_messages():
             assert msg.type in VALID_MESSAGE_TYPES
 
 
-def test_sigterm_generate_messages():
+def test_sigterm_generate_messages(tmp_path):
     """Test that sending sigterm to Trollflow2 stops it."""
     import os
     import signal
     from multiprocessing import Process
 
     from trollflow2.launcher import generate_messages
-
-    connection_parameters = {"nameserver": False, "addresses": "localhost:40000", "topic": "/test"}
+    ipc_path = os.fspath(tmp_path / "my_pipe")
+    connection_parameters = {"nameserver": False, "addresses": f"ipc://{ipc_path}", "topic": "/test"}
     proc = Process(target=generate_messages, args=(connection_parameters, ))
     proc.start()
     # Wait for the message listening loop to start
--- a/trollflow2/tests/test_launcher.py
+++ b/trollflow2/tests/test_launcher.py
@@ -887,5 +887,61 @@ def test_generate_messages():
             assert msg.type in VALID_MESSAGE_TYPES
 
 
+def test_sigterm_generate_messages():
+    """Test that sending sigterm to Trollflow2 stops it."""
+    import os
+    import signal
+    import time
+    from multiprocessing import Process
+
+    from trollflow2.launcher import generate_messages
+
+    connection_parameters = {"nameserver": False, "addresses": "localhost:40000", "topic": "/test"}
+    proc = Process(target=generate_messages, args=(connection_parameters, ))
+    proc.start()
+    tic = time.time()
+    # Wait for the message listening loop to start
+    time.sleep(1)
+    # Send SIGTERM
+    os.kill(proc.pid, signal.SIGTERM)
+    proc.join()
+
+    assert proc.exitcode == 0
+    # The queue.get timeout is set to 5 seconds, so it should be at
+    # least this long until the process is terminated
+    elapsed_time = time.time() - tic
+    assert elapsed_time >= 5.0
+
+
+def test_sigterm_runner(tmp_path):
+    """Test that sending sigterm to Trollflow2 stops it."""
+    import os
+    import signal
+    import time
+    from multiprocessing import Process
+
+    from trollflow2.launcher import Runner
+
+    product_list = tmp_path / "trollflow2.yaml"
+    with open(product_list, "w") as fid:
+        fid.write(yaml_test1)
+    connection_parameters = {"nameserver": False, "addresses": "localhost:40000", "topic": "/test"}
+    runner = Runner(product_list, connection_parameters)
+    proc = Process(target=runner.run)
+    proc.start()
+    tic = time.time()
+    # Wait for the message listening loop to start
+    time.sleep(1)
+    # Send SIGTERM
+    os.kill(proc.pid, signal.SIGTERM)
+    proc.join()
+
+    assert proc.exitcode == 0
+    # The queue.get timeout is set to 5 seconds, so it should be at
+    # least this long until the process is terminated
+    elapsed_time = time.time() - tic
+    assert elapsed_time >= 5.0
+
+
 if __name__ == '__main__':
     unittest.main()
--- a/trollflow2/tests/test_launcher.py
+++ b/trollflow2/tests/test_launcher.py
@@ -907,19 +907,33 @@ def test_sigterm_generate_messages():
     assert proc.exitcode == 0
 
 
-def test_sigterm_runner(tmp_path):
+def _fake_queue_logged_process(msg, prod_list, produced_files, **kwargs):
+    time.sleep(5.0)
+
+
+@mock.patch("trollflow2.launcher.ListenerContainer")
+@mock.patch("trollflow2.launcher.queue_logged_process",
+            new=_fake_queue_logged_process)
+def test_sigterm_runner(lc_, tmp_path, caplog):
     """Test that sending sigterm to Trollflow2 stops it."""
     import os
     import signal
-    import time
     from multiprocessing import Process
 
+    from posttroll.message import Message
+
     from trollflow2.launcher import Runner
+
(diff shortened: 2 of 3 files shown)


pnuu:
There are currently two tests. One running `generate_messages` in a subprocess, another that does the same with `Runner`. With the latter case I get an error with `zmq.error.ZMQError: Invalid argument (addr='localhost:40000')` eventhough the `connection_parameters` are equal to the former. This causes the exit code to be `1` instead of the expected `0`.

Another problem is, that I'd expect the process to be up for 5 seconds due to the message iteration loop to wait for 5 seconds for new messages before timeout. Instead the process terminates immediately.

pnuu:
If I test with this script
```python
#!/usr/bin/env python

import os
import signal
import time


def main():
    """"""
    print(os.getpid())
    keep_looping = True
    tic = None

    def _signal_handler(signum, frame):
        nonlocal keep_looping
        nonlocal tic
        print("signal received")
        keep_looping = False
        tic = time.time()
    
    signal.signal(signal.SIGTERM, _signal_handler)

    while keep_looping:
        time.sleep(10)

    print("terminated with", signal.SIGTERM)
    print("time after termination", time.time() - tic)


if __name__ == "__main__":
    main()
```

and run `kill` from another terminal I see the termination time to be in relation to the `time.sleep()`. I thought these should be pretty equivalent.

pnuu:
Also using
```python
    while keep_looping:
        try:
            _ = in_queue.get(True, 10)
        except queue.Empty:
            pass
```
works as expected and the 10 second timeout is respected.

pnuu:
I'm guessing the premature termination is because of the `posttroll.listener.ListenerContainer` runs in a separate thread and is killed so the queue also exits.

I'll see if I can adapt the `test_sigterm_runner` so that there is a proper "workload" running and that it runs until completion.

pnuu:
With https://github.com/pytroll/trollflow2/pull/217/commits/11e5dff668a3ecf6d66fb12c466d4b94f3fe8156 the processing should be simulated and shown that we are waiting for it to finish before exiting.

I'll add a note about the SIGTERM signaling to documentation.

trollflow2/tests/test_launcher.py
```diff
@@ -887,5 +887,69 @@ def test_generate_messages():
...
 
+def test_sigterm_generate_messages():
+    """Test that sending sigterm to Trollflow2 stops it."""
+    import os
+    import signal
+    from multiprocessing import Process
+
+    from trollflow2.launcher import generate_messages
+
+    connection_parameters = {"nameserver": False, "addresses": "localhost:40000", "topic": "/test"}
```
mraspaud, line 898: I've had problems to run tests with connections to localhost on some machines in the past. Would it be possible to switch this with an ipc connection instead?
eg:
```suggestion
def test_sigterm_generate_messages(tmp_path):
    """Test that sending sigterm to Trollflow2 stops it."""
    import os
    import signal
    from multiprocessing import Process

    from trollflow2.launcher import generate_messages
    ipc_path = os.fspath(tmp_path / "my_pipe")
    connection_parameters = {"nameserver": False, "addresses": f"ipc://{ipc_path}", "topic": "/test"}
```

trollflow2/tests/test_launcher.py
```diff
@@ -887,5 +887,69 @@ def test_generate_messages():
...
+    proc.join()
+
+    assert proc.exitcode == 0
+
+
+def _fake_queue_logged_process(msg, prod_list, produced_files, **kwargs):
+    time.sleep(5.0)
+
+
+@mock.patch("trollflow2.launcher.ListenerContainer")
```
mraspaud, line 914: Do we really need this mocking? would it work with posstroll.testing's `patched_subscriber_recv`? 
example usage: https://github.com/pytroll/pytroll-runner/blob/main/pytroll_runner/tests/test_runner.py#L323

mraspaud (commented):
Nice one, thanks for this. I have a couple of requests inline.
Then codecov is complaining about the signal_handler not being tested, but that can't be right, can it?

trollflow2/tests/test_launcher.py
pnuu: The connection is never used, it's basically just saying "don't bother connecting to a nameserver that isn't there". I don't mind changing to your suggestion but I don't understand what it does.

trollflow2/tests/test_launcher.py
pnuu: How does it work replacing `ListenerContainer`, the underlying `Listener` and the output queue?

doc/source/launcher.rst
```diff
@@ -11,3 +11,7 @@ messages by starting ``satpy_launcher.py`` with command-line arguments
 ``-n False -a tcp://<host>:<port>`` where the host and port point to a
 message publisher. Multiple publisher addresses can be given by supplying
 them with additional ``-a`` switches.
+
+The Trollflow2 process can be terminated gracefully by sending SIGTERM
+to the it, eg. by ``kill -SIGTERM <pid>`` where ``<pid>`` is the
+process ID number of Trollflow2.
```
mraspaud, line 17: One more thing: maybe we should explain that the gracefull termination implies finishing the current running processing?

pnuu: Clarified in https://github.com/pytroll/trollflow2/pull/217/commits/5ecd049b70863022b727d343dea5ca77ae65e135

trollflow2/tests/test_launcher.py
pnuu: I tried this:
```python
    with patched_subscriber_recv([msg]):
        runner = Runner(product_list, connection_parameters)
    proc = Process(target=runner.run)
    proc.start()
    ...
```

but it exited immediately with exitcode 15 (SIGTERM) after the `kill` command.

trollflow2/tests/test_launcher.py
mraspaud: ipc is inter-process communication and basically just uses a pipe file on linux. Ie, doesn't need a network to work. It's supported by pyzmq and thus posttroll, I use it in other tests. It's treated as just another protocol.

pnuu:
It is interesting that Codecov complains that `_signal_handler` is not covered, running all the tests locally with `pytest -s` prints out the signaling log message. I did try with `caplog` to catch those messages (the fixture seems to be still there in `test_sigterm_runner()`...), but I guess it didn't work because of the logs being in a separate thread/process and the messages are not propageted in my test case. I'll have another go.

pnuu:
Hmm, no, I can't capture the log messages even if I get them to print on the terminal. I also tried wrapping the `Process()` until `proc.join()` with `trollflow2.logging.logging_on` decorator but still only see the message on terminal and `caplog.text` stays empty.

pnuu:
And this is what is printed to terminal when using `with logging_on()` context:
```
[DEBUG: 2024-10-21 15:48:53 : trollflow2.launcher] Launching trollflow2 with subprocesses
[INFO: 2024-10-21 15:48:53 : trollflow2.launcher] New message received.
[DEBUG: 2024-10-21 15:48:53 : trollflow2.launcher] pytroll://my/topic file lahtinep@lahtinep21lk 2024-10-21T12:48:53.117334 v1.01 application/json {"filename": "foo"}
[INFO: 2024-10-21 15:48:54 : trollflow2.launcher] Caught signal to stop processing new messages and to terminate Trollflow2.
[INFO: 2024-10-21 15:48:59 : trollflow2.launcher] All 0 files produced nominally in 0:00:06.460422
```

trollflow2/tests/test_launcher.py
mraspaud: So `_fake_queue_logged_process`isn't called?

mraspaud:
> And this is what is printed to terminal when using `with logging_on()` context:
> 
> ```
> [DEBUG: 2024-10-21 15:48:53 : trollflow2.launcher] Launching trollflow2 with subprocesses
> [INFO: 2024-10-21 15:48:53 : trollflow2.launcher] New message received.
> [DEBUG: 2024-10-21 15:48:53 : trollflow2.launcher] pytroll://my/topic file lahtinep@lahtinep21lk 2024-10-21T12:48:53.117334 v1.01 application/json {"filename": "foo"}
> [INFO: 2024-10-21 15:48:54 : trollflow2.launcher] Caught signal to stop processing new messages and to terminate Trollflow2.
> [INFO: 2024-10-21 15:48:59 : trollflow2.launcher] All 0 files produced nominally in 0:00:06.460422
> ```

yes, caplog and the queue logger sometimes ignore each other. Question is in which thread/process does the signal handling run? That maybe influences where the log goes?

trollflow2/tests/test_launcher.py
pnuu: Yes, seems so, and the test completes in 1.007 seconds.

pnuu:
> Question is in which thread/process does the signal handling run? That maybe influences where the log goes?

In `test_sigterm_generate_messages()` the signal handling is in the main thread, there's only the one process.

It gets more complicated in `test_sigterm_runner()`. `generate_messages()` is called in `launcher.Runner._get_message_iterator()` and the iterator is passed on to `_run_threaded()` or `_run_subprocess()`. The `queue_logged_process` is given a message from the message receiving iterator. The message receiving part should thus be within the main process of `Runner` that is started in the test case.

trollflow2/tests/test_launcher.py
mraspaud: Ok, thinking more about this, I would pass two messages to the patched subscriber and check that only one is processed.
But that doesn't help with the fake process being ignored. The patching is still there, right?
I don't understand why this is not working, the patching of the subscriber only passes messages down. Unless the ListenerContainer bypasses it somehow? Can you check if the message is consumed?

pnuu: Ok, I think I figured it out. Things broke at https://github.com/pytroll/posttroll/blob/main/posttroll/listener.py#L47 because I had empty connection settings...

Switched to `patched_subscriber_recv` in https://github.com/pytroll/trollflow2/pull/217/commits/aefd2a33fd131f24065dcbafd123d968680c2406

But the `fake_queue_logged_process` is still not run. If I change it to sleep for 15 s the execution time doesn't change. But lets continue with that outside this thread.

pnuu:
Looks like the messages do not end up in the `listener.output_queue`. If I revert https://github.com/pytroll/trollflow2/pull/217/commits/aefd2a33fd131f24065dcbafd123d968680c2406 the `_fake_queue_logged_process` is called again.

mraspaud:
> Looks like the messages do not end up in the `listener.output_queue`. If I revert [aefd2a3](https://github.com/pytroll/trollflow2/commit/aefd2a33fd131f24065dcbafd123d968680c2406) the `_fake_queue_logged_process` is called again.

ok, then I suppose the patched subscriber doesn't work with the listenercontainer. That is another issue that we need to fix in another PR, so feel free to put the mocking back.

pnuu:
Ok, I reverted from `patched_subscriber_recv` to mocking `ListenerContainer`. I also shortened the output queue wait time from 5 to 1 second to make testing a bit more streamlined. And removed unused `caplog` from `test_sigterm_runner()`.

mraspaud (approved):
LGTM

PR #220, opened 2024-11-15
Fix log message printing the number of loaded composites

pnuu:
Trollflow2 is printing a log message saying `Loading 1 composites` even when there are more. This PR fixes the log message.

 - [x] Tests added  
 - [x] Tests passed 

### DIFF
--- a/trollflow2/tests/test_trollflow2.py
+++ b/trollflow2/tests/test_trollflow2.py
@@ -844,6 +844,11 @@ class TestCreateScene(TestCase):
 class TestLoadComposites(TestCase):
     """Test case for loading composites."""
 
+    @pytest.fixture(autouse=True)
+    def inject_fixtures(self, caplog):
+        """Inject fixtures."""
+        self._caplog = caplog
+
     def setUp(self):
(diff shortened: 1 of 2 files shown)


trollflow2/tests/test_trollflow2.py
```diff
@@ -844,6 +844,11 @@ def test_create_scene(self):
 class TestLoadComposites(TestCase):
     """Test case for loading composites."""
 
+    @pytest.fixture(autouse=True)
+    def inject_fixtures(self, caplog):
+        """Inject fixtures."""
+        self._caplog = caplog
```
gerritholl, line 850: What is the benefit of storing a `_caplog` attribute and using that, rather than using the `caplog` attribute in the tests directly?

gerritholl (approved):
Looks good to me, just one small question on caplog in the test.

trollflow2/tests/test_trollflow2.py
pnuu: This is the only way to use fixtures in `unittest.TestCase` classes.

PR #223, opened 2024-11-15
Fix trollsched usage when not installed

### LINKED ISSUE #222: The `check_valid_data_fraction` plugin does not check for pytroll-schedule availability
Out of the two other plugin that use the pytroll-schedule package, `covers()`, has the same check, but `check_valid_data_fraction()` does not. These are also using a public `get_scene_coverage()` function.

_Originally posted by @pnuu in https://github.com/pytroll/trollflow2/issues/132#issuecomment-2478204485_

pnuu:
The `check_valid_data_fraction()` doesn't have a check whether pytroll-schedule is installed or not.

This PR adds the check, and makes the `get_scene_coverage()` helper function a private function to show it should not be called directly.

 - [x] Closes #222 
 - [x] Tests updated
 - [x] Tests passed 

### DIFF
--- a/trollflow2/plugins/__init__.py
+++ b/trollflow2/plugins/__init__.py
@@ -639,8 +639,8 @@ def _check_overall_coverage_for_area(
     Helper for covers().
     """
     area_path = "/product_list/areas/%s" % area
-    cov = get_scene_coverage(platform_name, start_time, end_time,
-                             sensor, area)
+    cov = _get_scene_coverage(platform_name, start_time, end_time,
+                              sensor, area)
     product_list['product_list']['areas'][area]['area_coverage_percent'] = cov
     if cov < min_coverage:
         logger.info(
@@ -654,7 +654,7 @@ def _check_overall_coverage_for_area(
                      f"{min_coverage:.2f}% - Carry on with {area:s}")
 
 
-def get_scene_coverage(platform_name, start_time, end_time, sensor, area_id):
+def _get_scene_coverage(platform_name, start_time, end_time, sensor, area_id):
     """Get scene area coverage in percentages."""
     overpass = Pass(platform_name, start_time, end_time, instrument=sensor)
     area_def = get_area_def(area_id)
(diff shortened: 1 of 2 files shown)


trollflow2/plugins/__init__.py
```diff
@@ -968,6 +968,11 @@ def check_valid_data_fraction(job):
     """
     logger.info("Checking valid data fraction.")
 
+    if get_twilight_poly is None:
+        logger.error("Trollsched import failed, calculation of valid data fraction not possible")
+        logger.info("Keeping all products")
+        return
```
gerritholl, line 974: This should be possible to test by monkeypatching the value of `get_twilight_poly`, shouldn't it?

gerritholl (changes requested):
Please add a unit test for the added lines if possible.

trollflow2/plugins/__init__.py
pnuu: I need to research what monkeypatching means.

gerritholl: I was thinking [`monkeypatch.setattr`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.MonkeyPatch.setattr) should do the job and take care of restoring the original value after the test is completed.

pnuu: Yep, I'm looking at what the monkeypatch means and does.

pnuu: Test added in https://github.com/pytroll/trollflow2/pull/223/commits/07023a79121ae7d9a1f6e9156a38383b99cad84d

trollflow2/tests/test_trollflow2.py
```diff
@@ -2200,18 +2189,68 @@ def test_valid_filter(caplog, sc_3a_3b):
...
+
+    job, prods = _create_valid_filter_job_and_prods(sc_3a_3b)
+
+    with mock.patch("trollflow2.plugins._get_scene_coverage") as tpg, \
+            caplog.at_level(logging.DEBUG):
+        tpg.return_value = 100
+        check_valid_data_fraction(job)
+
+    assert "Trollsched import failed" in caplog.text
+    assert "Keeping all products" in caplog.text
```
pnuu, line 2237: This seems to work only if it is run by it self. If all the tests are run, the tests fails.

pnuu: Fixed in https://github.com/pytroll/trollflow2/pull/223/commits/4058756893421c247ad399895a6cf23cc9a7ecd6

mraspaud (approved):
LGTM! thanks for fixing this check and making the code better along the way!

PR #224, opened 2024-11-15
Put output dir to tmp_path

### LINKED ISSUE #155: Test directory not removed after tests
**Describe the bug**
A directory created during unit tests is not deleted.

**To Reproduce**

```
pytest .
```

**Expected behavior**
I would like the directory to be deleted

**Actual results**
A directory called `örülök, hogy megismerhetem` is left in the directory where the test was run.

A good solution would probably be to use pytest's `tmp_path` fixture.

@gerritholl I think that's one of the tests you added, right?

pnuu:
The `TestSaveDatasets.test_pop_unknown_args` is creating a Hungarian directory and not removing it. This PR fixes it by using `tmp_path` fixture.

 - [x] Closes #155 
 - [x] Tests modified
 - [x] Tests passed 


gerritholl (approved):
I didn't think with all the mocking anything would be created at all, but looks good to me.

PR #229, opened 2025-04-07
Migrate to pyproject.toml

mraspaud:
This PR replaces setup.py with pyproject, and uses hatchling as a build system


pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+[project]
+name = "trollflow2"
+dynamic = ["version"]
+description = "Pytroll workflow execution framework"
+readme = "README.md"
+license = { text = "GPLv3+" }
+requires-python = ">=3.9"
+authors = [
+    { name = "Martin Raspaud", email = "martin.raspaud@smhi.se" },
+]
```
pnuu, line 14: Should there be more authors listed?

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+    "dpath>=2.1.0",
+    "posttroll>=1.10.0",
+    "pyorbital",
+    "pyyaml",
+    "satpy>=0.32.0",
+    "trollsift",
+]
+
+[project.scripts]
+satpy_cli = "trollflow2.cli:cli"
```
pnuu, line 33: `satpy_launcher.py` needs to be included (and removed from `bin/`).

pnuu (changes requested):

pyproject.toml
mraspaud: Absolutely, I just did an automatic migration with hatch for the most, didn't look too much at the metadata sorry

PR #240, opened 2025-09-12
Fix automatic upload of new releases

mraspaud:
Release of 0.17 automatic upload did not go through sdist since setup.py was removed.


mraspaud:
> 🤦‍♂️ I should have caught this.

No worries.

For the record, I uploaded the dist packages manually, so there is no rush in making a new release.

PR #241, opened 2025-09-24
Fix version where area and writer imports were changed

pnuu:
While testing for upcoming processing updates, I noticed that I still got warnings about some imports being from the "old" Satpy packages. The reason was that the Satpy version number was set to `0.59.0` while the changes were already in `0.58.0`.

 - [ ] Closes #xxxx 
 - [ ] Tests added  
 - [ ] Tests passed 
 - [ ] Fully documented 


mraspaud (approved):
LGTM

PR #244, opened 2026-01-20
Support python 3.14

mraspaud:
With new handling of multiprocessing and forking, python 3.14 broke a few things in trollflow2.

This PR fixes it.

 - [ ] Tests passed


pnuu (approved):
LGTM.

So the `Manager` didn't work when it was a global varible, and the cache usage fixes it?

